PR #9, opened 2020-09-13
Homework - unique_ptr

anraf1001:


jakjan95 (approved):
Code looks very good, just add delete statement for copy assignment and it will be perfect :)

anraf1001:
> Code looks very good, just add delete statement for copy assignment and it will be perfect :)

Copy constructor has been already deleted so copying unique_ptr isn't allowed

homework/unique_ptr/PtrIsNullptr.hpp
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include <stdexcept>
+#include <string>
+
+class PtrIsNullptr : public std::runtime_error {
+public:
+    PtrIsNullptr()
+        : std::runtime_error("Pointer is nullptr!!!") {}
```
MaciejSalamonski, line 9: +1

MaciejSalamonski (approved):
Very nice code! :}

homework/unique_ptr/unique_ptr.hpp
```diff
@@ -0,0 +1,61 @@
...
+    return temp;
+}
+
+template <typename T>
+T& unique_ptr<T>::operator*() const {
+    if (!ptr_) {
+        throw PtrIsNullptr{};
+    }
+    return *ptr_;
+}
```
ziobron, line 61: you have an additional exception in case of nullptr dereference. Casual unique_ptr does not check it, but it's ok for me for this implementation.

ziobron (changes requested):
An additional CI job with Valgrind would prove that there are no memory leaks. I think I have discovered one. Please check it.

ziobron (approved):

ziobron:

PR #14, opened 2020-09-15
unique_ptr homework

Kuba-Wi:


homework/unique_ptr/unique_ptr.hpp
```diff
@@ -0,0 +1,59 @@
+#pragma once
+
+template<typename T>
+class unique_ptr {
+public:
+    unique_ptr() = default;
+    unique_ptr(T* pointer) : pointer_(pointer) {}
+    unique_ptr(unique_ptr& ptr);
+    unique_ptr(unique_ptr&& ptr);
+    ~unique_ptr();
```
anraf1001, line 10: You should delete copy constructor to disable copying unique_ptrs

homework/unique_ptr/unique_ptr.hpp
```diff
@@ -0,0 +1,59 @@
...
+    void reset(T* new_ptr);
+private:
+    T* pointer_ = nullptr;
+};
+
+template<typename T>
+unique_ptr<T>::~unique_ptr() {
+    if(pointer_ != nullptr) {
+        delete pointer_;
+        pointer_ = nullptr;
```
anraf1001, line 24: You don't have to make `pointer_` nullptr. After dtor execution that whole object will be removed from memory.

anraf1001 (commented):
Some comments

homework/unique_ptr/unique_ptr.hpp
Kuba-Wi: It's better to set nullptr because later it is easier to check if this pointer points on something.

homework/unique_ptr/unique_ptr.hpp
Kuba-Wi: Actually they move pointer and don't copy.

homework/unique_ptr/unique_ptr.hpp
anraf1001: But we are talking about destructor. That whole object won't exist "in a second".

homework/unique_ptr/unique_ptr.hpp
anraf1001: Ok I haven't seen that you've created copy and move constructor that do exactly the same thing. I still think that you should delete copy constructor instead of making it pseudo move ctor but this's is only my point of view.

homework/unique_ptr/unique_ptr.hpp
MaciejSalamonski: I agree with anraf

https://stackoverflow.com/questions/3116980/do-i-have-to-set-pointer-to-nullptr-in-destructor

MaciejSalamonski (approved):
good job  👍

homework/unique_ptr/.github/module3.yml
```diff
@@ -0,0 +1,20 @@
+name: Homework unique_ptr
+
+on: [push, pull_request]
+
+jobs:
+  unique_ptr:
```
ziobron, line 6: Please add a valgrind / sanitizer job to prove that there are no memory leaks

homework/unique_ptr/test.cpp
```diff
@@ -0,0 +1,41 @@
...
+    unique_ptr<int> ptr(new int{42});
+    *ptr = 1;
+
+    ASSERT_TRUE(*ptr == 1);
+}
+
+TEST(UniquePtrTest, ArrowOperatorShouldReturnObject) {
+    unique_ptr<std::string> ptr(new std::string{"word"});
+
+    ASSERT_TRUE(ptr->size() == 4);
```
ziobron, line 16: 4 is a magic value. Avoid it.

homework/unique_ptr/test.cpp
```diff
@@ -0,0 +1,41 @@
...
+}
+
+TEST(UniquePtrTest, GetFunctionShouldReturnRawPointerAndResetChangePointer) {
+    unique_ptr<int> ptr;
+    ASSERT_TRUE(ptr.get() == nullptr);
+
+    ptr.reset(new int{42});
+    ASSERT_TRUE(ptr.get() != nullptr);
+    ASSERT_TRUE(*ptr.get() == 42);
```
ziobron, line 25: Use ASSERT_EQ / ASSERT_NE

homework/unique_ptr/test.cpp
```diff
@@ -0,0 +1,41 @@
...
+}
+
+TEST(UniquePtrTest, CopyingConstructorShouldSetNullptr) {
+    unique_ptr<int> temp(new int{42});
+    unique_ptr<int> main_ptr(temp);
+
+    ASSERT_TRUE(temp.get() == nullptr);
+    ASSERT_TRUE(main_ptr.get() != nullptr);
```
ziobron, line 33: Copy operations should be prohibited.

homework/unique_ptr/test.cpp
```diff
@@ -0,0 +1,41 @@
...
+}
+
+TEST(UniquePtrTest, ReleaseFunctionShouldSetNullptrAndReturnPointer) {
+    unique_ptr<int> temp(new int{42});
+    int* ptr = temp.release();
+
+    ASSERT_TRUE(*ptr == 42);
+}
```
ziobron, line 41: you should also delete ptr here.

homework/unique_ptr/unique_ptr.hpp
```diff
@@ -0,0 +1,59 @@
...
+    }
```
ziobron, line 25: just
```suggestion
    delete pointer_;
```
It's safe to delete nullptr.

homework/unique_ptr/unique_ptr.hpp
ziobron: @anraf1001 is right. Now you have created an auto_ptr which is very easy to use incorrectly. For example getting an element from vector (via operator[]) will invalidate the element in collection. Definitely disable the usage of copy operations

ziobron, line 10: You should also have a move and copy assignment operators.

homework/unique_ptr/unique_ptr.hpp
```diff
@@ -0,0 +1,59 @@
...
+    unique_ptr() = default;
+    unique_ptr(T* pointer) : pointer_(pointer) {}
+    unique_ptr(unique_ptr& ptr);
+    unique_ptr(unique_ptr&& ptr);
+    ~unique_ptr();
+    T& operator*() const { return *pointer_; }
+    T* operator->() const { return pointer_; }
+    T* get() const { return pointer_; }
+    T* release();
+    void reset(T* new_ptr);
```
ziobron, line 15:
```suggestion
    void reset(T* new_ptr = nullptr);
```
The default value will be useful here

homework/unique_ptr/unique_ptr.hpp
```diff
@@ -0,0 +1,59 @@
...
+    T* temp = pointer_;
+    pointer_ = nullptr;
+    return temp;
+}
+
+template<typename T>
+void unique_ptr<T>::reset(T* new_ptr) {
+    if(pointer_ != nullptr) {
+        delete pointer_;
+    }
```
ziobron, line 39: `if` not needed.
Just `delete pointer_;`
Deleting nullptr is safe.

homework/unique_ptr/unique_ptr.hpp
```diff
@@ -0,0 +1,59 @@
...
+    pointer_ = ptr.pointer_;
+    ptr.pointer_ = nullptr;
+}
+
+template<typename T>
+unique_ptr<T>::unique_ptr(unique_ptr&& ptr) {
+    if(pointer_ != nullptr) {
```
ziobron, line 54: if not needed.

homework/unique_ptr/unique_ptr.hpp
```diff
@@ -0,0 +1,59 @@
...
+}
+
+template<typename T>
+unique_ptr<T>::unique_ptr(unique_ptr& ptr) {
+    if(pointer_ != nullptr) {
+        delete pointer_;
+    }
+    pointer_ = ptr.pointer_;
+    ptr.pointer_ = nullptr;
+}
```
ziobron, line 50: prohibit using a copy constructor by = delete;
This behaviour is incorrect.

homework/unique_ptr/test.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+TEST(UniquePtrTest, CopyingConstructorShouldSetNullptr) {
+    unique_ptr<int> temp(new int{42});
+    unique_ptr<int> main_ptr(temp);
+
+    ASSERT_TRUE(temp.get() == nullptr);
+    ASSERT_TRUE(main_ptr.get() != nullptr);
+}
+
```
ziobron, line 35: there are no tests for move operation

ziobron (changes requested):
Use constexpr for magic values.

WBigaj (commented):

homework/unique_ptr/test.cpp
```diff
@@ -27,15 +27,37 @@ TEST(UniquePtrTest, GetFunctionShouldReturnRawPointerAndResetChangePointer) {
...
+TEST(UniquePtrTest, CopyAssignOperatorShouldAssignWhenMainPtrIsNotNullptr) {
+    unique_ptr<int> temp(new int{42});
+    unique_ptr<int> main_ptr(new int{1});
+    main_ptr = std::move(temp);
+
+    ASSERT_TRUE(temp.get() == nullptr);
+    ASSERT_TRUE(main_ptr.get() != nullptr);
+    ASSERT_TRUE(*main_ptr == 42);
```
ziobron, line 54: 42 is a magic value. Name it and reuse it.

homework/unique_ptr/test.cpp
```diff
@@ -27,15 +27,37 @@ TEST(UniquePtrTest, GetFunctionShouldReturnRawPointerAndResetChangePointer) {
...
+}
+
+TEST(UniquePtrTest, CopyAssignOperatorShouldSetNullptr) {
+    unique_ptr<int> temp(new int{42});
+    unique_ptr<int> main_ptr;
+    main_ptr = std::move(temp);
+
+    ASSERT_TRUE(temp.get() == nullptr);
+    ASSERT_TRUE(main_ptr.get() != nullptr);
+    ASSERT_TRUE(*main_ptr == 42);
```
ziobron, line 44: Instead of ASSERT_TRUE you should use ASSERT_EQ and ASSERT_NE

homework/unique_ptr/unique_ptr.hpp
```diff
@@ -3,18 +3,19 @@
...
+    unique_ptr& operator=(unique_ptr&& ptr);
     ~unique_ptr();
     T& operator*() const { return *pointer_; }
     T* operator->() const { return pointer_; }
     T* get() const { return pointer_; }
     T* release();
     void reset(T* new_ptr = nullptr);
 private:
-    T* pointer_ = nullptr;
+    T* pointer_;
```
ziobron, line 18:
```suggestion
    T* pointer_ = nullptr;
```
The previous version was better.

ziobron (approved):

ziobron:
3 XP za punktualność

PR #16, opened 2020-09-16
Homework - UniquePtr

MonikaZelechowska:


jakjan95 (approved):
Everything looks good! :)

MaciejSalamonski (commented):
Hi! You have sent two implementations of unique_ptr. Which implementation should I check?

homework/unique_ptr/UniquePtr-ut.cpp
```diff
@@ -0,0 +1,45 @@
...
+TEST_F(UniquePtrTest, ShouldUninitializedUniquePtrBeNullPtr) {
+    UniquePtr<int> uniqueNullPtr{};
+    ASSERT_EQ(uniqueNullPtr.get(), nullptr);
+}
+
+TEST_F(UniquePtrTest, ShouldReleaseUniquePtr) {
+    auto tempPtr = testPtr.release();
+    ASSERT_EQ(*tempPtr, initValue);
+    ASSERT_EQ(testPtr.get(), nullptr);
+}
```
ziobron, line 33: don't forget to `delete tempPtr;`

homework/unique_ptr/UniquePtr-ut.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+TEST_F(UniquePtrTest, ShouldResetUniquePtr) {
+    int* tempPtr = new int{newValue};
+    testPtr.reset(tempPtr);
+    ASSERT_EQ(*testPtr, newValue);
+}
```
ziobron, line 39: also a reset without an argument should be tested
`testPtr.reset();`

homework/unique_ptr/UniquePtr-ut.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+TEST_F(UniquePtrTest, ShouldMoveUniquePtr) {
+    auto tempPtr(std::move(testPtr));
+    ASSERT_EQ(*tempPtr, initValue);
+    ASSERT_EQ(testPtr.get(), nullptr);
+}
```
ziobron, line 45: also moving to a pointer that already holds an object should be tested.

exercises/uniquPointer.cpp
```diff
@@ -0,0 +1,164 @@
+#include <iostream>
```
ziobron, line 1: I'm not checking this implementation because it's in another path.

homework/unique_ptr/UniquePtr.hpp
```diff
@@ -0,0 +1,91 @@
...
+    ~UniquePtr();
+    UniquePtr(const UniquePtr& otherPtr) = delete;
+    UniquePtr(UniquePtr&& otherPtr);
+    UniquePtr& operator=(const UniquePtr& otherPtr) = delete;
+    UniquePtr& operator=(UniquePtr&& otherPtr);
+    T& operator*() const;
+    T* operator->() const;
+    T* get() const;
+    T* release();
+    void reset(T* newPtr);
```
ziobron, line 26:
```suggestion
    void reset(T* newPtr = nullptr);
```
default argument would be useful :)

homework/unique_ptr/UniquePtr.hpp
```diff
@@ -0,0 +1,91 @@
...
+}
+
+template <typename T>
+UniquePtr<T>::UniquePtr(UniquePtr<T>&& otherPtr)
+    : ptr_(otherPtr.release()) {}
+
+template <typename T>
+UniquePtr<T>& UniquePtr<T>::operator=(UniquePtr<T>&& otherPtr) {
+    if (this != &otherPtr) {
+        if (ptr_ != nullptr) {
```
ziobron, line 48: this if is not necessary

homework/unique_ptr/UniquePtr.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+template <typename T>
+T& UniquePtr<T>::operator*() const {
+    try {
+        if (ptr_ == nullptr) {
+            throw DereferenceNullPtrError();
+        }
+    } catch (DereferenceNullPtrError& error) {
+        std::cerr << error.what() << "\n";
+    }
```
ziobron, line 65: If we use exceptions then you probably don't want to catch it here. It should propagate and it should be caught somewhere else, eg. in tests by ASSERT_THROW

homework/unique_ptr/UniquePtr.hpp
```diff
@@ -0,0 +1,91 @@
...
+template <typename T>
+T* UniquePtr<T>::release() {
+    T* tempPtr = ptr_;
+    ptr_ = nullptr;
+    return tempPtr;
+}
+
+template <typename T>
+void UniquePtr<T>::reset(T* newPtr) {
+    if (ptr_ != nullptr) {
```
ziobron, line 87: if not necessary

ziobron (changes requested):
Nice solution :)
A few improvements will be useful

homework/unique_ptr/UniquePtr.hpp
MonikaZelechowska: Thanks for advice. I used to use _try, catch _and_ throw_ in wrong way :)

homework/unique_ptr/UniquePtr-ut.cpp
```diff
@@ -30,16 +36,32 @@ TEST_F(UniquePtrTest, ShouldReleaseUniquePtr) {
...
     ASSERT_EQ(*tempPtr, initValue);
     ASSERT_EQ(testPtr.get(), nullptr);
+    delete tempPtr;
 }
 
 TEST_F(UniquePtrTest, ShouldResetUniquePtr) {
     int* tempPtr = new int{newValue};
     testPtr.reset(tempPtr);
     ASSERT_EQ(*testPtr, newValue);
+    delete tempPtr;
```
ziobron, line 46: you should not call delete here. If reset does swap it will work, however, reset should only copy a source pointer and not modify it.

homework/unique_ptr/UniquePtr-ut.cpp
```diff
@@ -30,16 +36,32 @@ TEST_F(UniquePtrTest, ShouldReleaseUniquePtr) {
...
+}
+
+TEST_F(UniquePtrTest, ShouldResetUniquePtrWithNoArgument) {
+    int* tempPtr = new int{newValue};
+    testPtr.reset();
+    ASSERT_EQ(testPtr.get(), nullptr);
+    delete tempPtr;
```
ziobron, line 53: the same here - without delete.

ziobron (changes requested):
Almost good :)
It would be good to see that it works on CI with 2 jobs - casual UTs and UTs under valgrind or compiled with ASAN. It would prove that everything works fine.

ziobron:
The code looks good, but I still need a proof - a working CI with valgrind / ASAN :)

WBigaj (commented):
Why have you sent two versions of program?

ziobron (approved):

ziobron:
3 XP za punktualność

PR #21, opened 2020-09-18
unique_ptr homework

McKucia:


ziobron (changes requested):
There are no unit tests. Without them you cannot prove that you have a correct implementation. Also you need to run them under valgrind or compile them with asan.

module1/homework/unique_ptr.cpp
```diff
@@ -0,0 +1,46 @@
+#include <iostream>
+
+template <class T>
+class unique_ptr{
+public:
+//    unique_ptr(T& ptr) : ptr_(&ptr) {};
```
WBigaj, line 6: What is this comment?

WBigaj (commented):
I have checked your code and have one question.

module1/homework/unique_ptr.hpp
```diff
@@ -0,0 +1,48 @@
...
+        return *ptr_;
+    }
+
+    T* release(){
+        T* tmpPtr = ptr_;
+        ptr_ = nullptr;
+        return tmpPtr;
+    }
+
+    void reset(T* otherPtr){
```
ziobron, line 41: reset can also be called without an argument. If you add a default argument value `T* otherPtr = nullptr` it will cover this case. ALso add a test for it.

ziobron (changes requested):
Looks good, but please add a CI job to prove that there are no memory leaks under valgrind. Remember to add running a valgrind job.

PR #23, opened 2020-09-19
Homework - Unique_ptr

Pawellos:
Implementation own Unique_ptr (constructor, move constructor, destructor, operatos, methods: get, reset, release, removed copy constructor)


homework/unique_ptr/UniquePointer.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+template <class T>
+UniquePointer<T>::UniquePointer() {}
+
+template <class T>
+UniquePointer<T>::UniquePointer(T* ptr) : ptr_(ptr) {}
+
+template <class T>
+UniquePointer<T>::~UniquePointer() {
+    if (ptr_) {
```
MaciejSalamonski, line 38: I think that You don't have to check if ptr isn't null pointer in destructor

https://stackoverflow.com/questions/50569942/check-pointer-is-not-null-in-destructor

Delete or delete[] are defined for nullptr. They do nothing

MaciejSalamonski (approved):
Hi nice code! Have a great day ;~}

homework/unique_ptr/UniquePointer.hpp
Pawellos: Ok, thanks. I was not sure :)

homework/unique_ptr/UniquePointer.hpp
```diff
@@ -0,0 +1,90 @@
...
+    UniquePointer();
+    ~UniquePointer();
+    UniquePointer(const UniquePointer&) = delete;
+    UniquePointer(UniquePointer&& anotherPtr);
+
+    T* get();
+    T* release();
+    void reset(T* ptr);
+
+    T operator*() const;
```
ziobron, line 21:
```suggestion
    T& operator*() const;
```
This operator should return a reference, not a copy

homework/unique_ptr/UniquePointer.hpp
```diff
@@ -0,0 +1,90 @@
...
+    T* operator->() const;
+    UniquePointer<T>& operator=(UniquePointer<T>&& anotherPtr);
+    UniquePointer<T>& operator=(const UniquePointer<T>&) = delete;
+
+private:
+    T* ptr_;
+};
+
+template <class T>
+UniquePointer<T>::UniquePointer() {}
```
ziobron, line 31: default constructor has uninitialized pointer, so it causes undefined behavior ❗️

homework/unique_ptr/test.cpp
```diff
@@ -0,0 +1,65 @@
...
+struct TestUniquePtr : ::testing::Test {
+    TestUniquePtr() : uPtr(new int{testValue1}) {}
+    UniquePointer<int> uPtr;
+};
+
+TEST_F(TestUniquePtr, TestOfConstructor) {
+    ASSERT_EQ(testValue1, *uPtr);
+
+    UniquePointer<int> ptr{new int{33}};
+    ASSERT_EQ(33, *ptr);
```
ziobron, line 17: avoid magic values. Name them.

homework/unique_ptr/test.cpp
```diff
@@ -0,0 +1,65 @@
...
+    UniquePointer<int> ptr(new int(15));
+    ptr.reset(new int(45));
+    ASSERT_EQ(45, *ptr);
+}
+
+TEST_F(TestUniquePtr, TestOfReleaseMethod) {
+    UniquePointer<int> ptr(new int(35));
+    ptr.release();
+    ASSERT_EQ(ptr.get(), nullptr);
+}
```
ziobron, line 42: there is a memory leak in this test case. You should assign a result of release() function to a pointer and delete it.

homework/unique_ptr/test.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+TEST_F(TestUniquePtr, TestOfNullPtrException) {
+    UniquePointer<int> ptr{nullptr};
+    ASSERT_THROW(*ptr, NullPtrException);
+}
+
+TEST_F(TestUniquePtr, TestOfMoveConstructor) {
+    UniquePointer<int> ptr(std::move(uPtr));
+    ASSERT_EQ(20, *ptr);
```
ziobron, line 51: What is 20? Why 20? Magic value.
What about uPtr after move? Also check if it's in a valid state.

homework/unique_ptr/test.cpp
```diff
@@ -0,0 +1,65 @@
...
+}
+
+TEST_F(TestUniquePtr, TestOfMoveAssignmentOperator) {
+    UniquePointer<int> ptr{new int{14}};
+    UniquePointer<int> ptr2 = std::move(ptr);
+    ASSERT_EQ(14, *ptr2);
```
ziobron, line 57: what about ptr? you should also test it.

ziobron (changes requested):

ziobron:
And it will be useful to have a CI job with valgrind / asan to prove that there are no memory leaks.

WBigaj (commented):

ziobron:
Ok, but please move homework/unique_ptr/.github/workflows/unique.yml to .github/workflows/unique.yml. It will trigger CI and I will be able to check results.

Pawellos:
> Ok, but please move homework/unique_ptr/.github/workflows/unique.yml to .github/workflows/unique.yml. It will trigger CI and I will be able to check results.

Ok, I think it's done  :)

ziobron:
Invalid workflow file
You have an error in your yaml syntax on line 7

Probably too many spaces in indentation. 4 instead of 2. Check it.

.github/workflows/unique.yml
```diff
@@ -0,0 +1,38 @@
+name: C++ with CMake CI
+
+on: [push, pull_request]
+
+jobs:
+  UniquePtr:
+      runs-on: ubuntu-20.04
```
ziobron, line 7:
```suggestion
    runs-on: ubuntu-20.04
```

.github/workflows/unique.yml
```diff
@@ -34,5 +34,5 @@ jobs:
     - name: Run valgrind 
       working-directory: homework/unique_ptr/build
       run: |
-        valgrind ./UniquePtr-ut
-        valgrind ./UniquePtr
+        valgrind ./test
```
ziobron, line 37: no such file or directory ./test. How is your UT binary named?

Pawellos: I am lost with this. I think it is test.cpp

ziobron: You can get this from CMakeLists.txt

```cmake
project(unique_ptr)
 enable_testing()

 add_executable(${PROJECT_NAME} main.cpp)
 add_executable(${PROJECT_NAME}-ut test.cpp)
```

It is `${PROJECT_NAME}-ut`, so it's `unique_ptr-ut`

Pawellos: Thanks ! I was messing with this a while and i was not sure what is wrong. Anyway thanks for help :)

ziobron:
3 XP za punktualność

PR #24, opened 2020-09-19
Unique ptr homework

niemarcin:


homework/unique_ptr/unique_ptr.hpp
```diff
@@ -0,0 +1,87 @@
+#pragma once
+
+namespace coders {
```
MaciejSalamonski, line 3: +1

homework/unique_ptr/unique_ptr.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+template <typename T>
+unique_ptr<T>::unique_ptr(unique_ptr<T>&& ptr) noexcept {
+    rawPtr_ = ptr.rawPtr_;
+    ptr.rawPtr_ = nullptr;
+}
+
+template <typename T>
+unique_ptr<T>::~unique_ptr() {
+    if (rawPtr_) {
```
MaciejSalamonski, line 41: I think that You don't have to check if pointer is nullpointer in destructor

https://stackoverflow.com/questions/50569942/check-pointer-is-not-null-in-destructor

Delete and delete[] are defined for nullptr. They do nothing

MaciejSalamonski (approved):
Very nice code! Keep it up ;~} Have a great day!!

homework/unique_ptr/unique_ptr-ut.cpp
```diff
@@ -0,0 +1,110 @@
...
+}
+
+TEST(UniquePointerTest, ShouldReleaseUniquePointerToString)
+{
+    // Given
+    coders::unique_ptr<std::string> uniquePtr(new std::string("Hello world!"));
+
+    // When
+    auto rawPtrBeforeRelease = uniquePtr.get();
+    auto releasedRawPtr = uniquePtr.release();
```
ziobron, line 38: there is a memory leak. You should delete releasedRawPtr at the end of this test case.

homework/unique_ptr/unique_ptr-ut.cpp
```diff
@@ -0,0 +1,110 @@
...
+{
+    // Given
+    int testValue = 10;
+    coders::unique_ptr<int> ptr1(new int(testValue));
+
+    // When
+    coders::unique_ptr<int> ptr2(std::move(ptr1));
+
+    // Then
+    ASSERT_EQ(*ptr2, testValue);
```
ziobron, line 68: you should also check ptr1 after move.

homework/unique_ptr/unique_ptr-ut.cpp
```diff
@@ -0,0 +1,110 @@
...
+    // Given
+    double testValue = 66.6;
+    coders::unique_ptr<double> ptr1(new double(testValue));
+    coders::unique_ptr<double> ptr2(new double(123.45));
+
+    // When
+    ptr2 = std::move(ptr1);
+
+    // Then
+    ASSERT_EQ(*ptr2, testValue);
```
ziobron, line 82: check ptr1 as well

homework/unique_ptr/unique_ptr-ut.cpp
```diff
@@ -0,0 +1,110 @@
...
+{
+    // Given
+    int testValue = 10;
+    coders::unique_ptr<int> ptr1(new int(testValue));
+
+    // When
+    coders::unique_ptr<int> ptr2(std::move(ptr1));
+
+    // Then
+    ASSERT_EQ(ptr1.get(), nullptr);
```
ziobron, line 95: ok, here is this check. Merge it into one test case.

homework/unique_ptr/unique_ptr-ut.cpp
```diff
@@ -0,0 +1,110 @@
...
+    // Given
+    int testValue = 10;
+    coders::unique_ptr<int> ptr1(new int(testValue));
+    coders::unique_ptr<int> ptr2(new int(666));
+
+    // When
+    ptr2 = std::move(ptr1);
+
+    // Then
+    ASSERT_EQ(ptr1.get(), nullptr);
```
ziobron, line 109: the same here :)

homework/unique_ptr/unique_ptr-ut.cpp
```diff
@@ -0,0 +1,110 @@
...
+    ASSERT_EQ(uniquePtr.get(), nullptr);
+}
+
+TEST(UniquePointerTest, ShouldResetUniquePointerToVectorOfInts)
+{
+    // Given
+    coders::unique_ptr<std::vector<int>> uniquePtr(new std::vector<int>{1, 2, 3});
+
+    // When
+    uniquePtr.reset(new std::vector<int>{10, 11});
```
ziobron, line 51: i'm missing a test case on reset without argument - `uniquePtr.reset();`

homework/unique_ptr/unique_ptr.hpp
```diff
@@ -0,0 +1,87 @@
+#pragma once
+
+namespace coders {
+
+template <typename T>
+class unique_ptr{
+public:
+    unique_ptr() = delete;
```
ziobron, line 8: why don't we want to create a unique_ptr like this:
`unique_ptr<int> uptr {};` 
It's possible with std::unique_ptr.

homework/unique_ptr/unique_ptr.hpp
```diff
@@ -0,0 +1,87 @@
...
+    T* tempPtr = rawPtr_;
+    rawPtr_ = nullptr;
+    return tempPtr;
+}
+
+template <typename T>
+void unique_ptr<T>::reset(T* ptr) noexcept {
+    T* oldPtr = rawPtr_;
+    rawPtr_ = ptr;
+    if (oldPtr) {
```
ziobron, line 62: if not necessary. It's safe to delete nullptr.

homework/unique_ptr/unique_ptr.hpp
```diff
@@ -0,0 +1,87 @@
...
+public:
+    unique_ptr() = delete;
+    unique_ptr(T* ptr);
+    unique_ptr(const unique_ptr<T>&) = delete;
+    unique_ptr(unique_ptr<T>&& ptr) noexcept;
+    ~unique_ptr();
+
+    T* get() const noexcept;
+    T* release() noexcept;
+    void reset(T* ptr) noexcept;
```
ziobron, line 16:
```suggestion
    void reset(T* ptr = nullptr) noexcept;
```
Default argument would be useful

ziobron (changes requested):

homework/unique_ptr/unique_ptr.hpp
mariuszlisowski: https://en.cppreference.com/w/cpp/language/delete
"If expression evaluates to a null pointer value, no destructors are called, and the deallocation function may or may not be called (it's unspecified), but the default deallocation functions are guaranteed to do nothing when passed a null pointer."

Well, it's unspecified as mentioned above. In my case I have had errors of double deleted pointer unless checked if it had been nullptr.

mariuszlisowski: Maybe I miss something...

WBigaj (commented):

ziobron (approved):

ziobron:
3 XP za punktualność

PR #32, opened 2020-09-29
zarzadzanie zasobami #1 solution

marcindus:
My interpretation of unique_ptr. I was trying to avoid looking at solution from
Zarządzanie zasobami #2 ..  So propably i can improve this solution.


homework/unique_ptr/test.cpp
```diff
@@ -0,0 +1,106 @@
...
+*/
+
+/*
+
+TEST(Test, deletedCopyConstructor_shouldNotCompile_withDeletedMethodError)
+{
+    my_unique_ptr<int> ptr  = my_unique_ptr<int>(new int(10));
+    my_unique_ptr<int> copy_ptr = my_unique_ptr<int>(ptr);
+}
+*/
```
WBigaj, line 106: What is that commentS?

WBigaj (commented):
I have checked your code and have one question.

marcindus:
I've created some test which will not compile since copy constructor and operator= are deleted in my_unique_ptr, i do not know how to test situation when compilation should fail

homework/unique_ptr/my_unique_ptr.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+    void reset(T* new_ptr)
+    {
+        delete ptr_;
+        ptr_ = new_ptr;
+    };
+
+    ~my_unique_ptr() { delete ptr_; }
+
+    T operator*() { return *ptr_; };
```
MaciejSalamonski, line 35: I think that return type should be T&. What if ptr is a nullptr?

homework/unique_ptr/my_unique_ptr.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+    my_unique_ptr() { ptr_ = nullptr; };
+
+    my_unique_ptr(const my_unique_ptr<T>&) = delete;
+
+    my_unique_ptr(my_unique_ptr<T>&& ptr_m) { ptr_ = ptr_m.release(); }
+
+    void operator=(const my_unique_ptr<T>&) = delete;
+
+    T* get() { return ptr_; }
```
MaciejSalamonski, line 18: Maybe get() const?

homework/unique_ptr/my_unique_ptr.hpp
```diff
@@ -0,0 +1,40 @@
...
+public:
+    explicit my_unique_ptr(T* ptr) : ptr_(ptr){};
+
+    my_unique_ptr() { ptr_ = nullptr; };
+
+    my_unique_ptr(const my_unique_ptr<T>&) = delete;
+
+    my_unique_ptr(my_unique_ptr<T>&& ptr_m) { ptr_ = ptr_m.release(); }
+
+    void operator=(const my_unique_ptr<T>&) = delete;
```
MaciejSalamonski, line 16: Where is operator=(&&)?

homework/unique_ptr/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
+cmake_minimum_required(VERSION 3.5)
```
MaciejSalamonski, line 1: I think you should add a workflow file (should contain valgrind) to prove that there are no memory leaks

MaciejSalamonski (approved):

jakjan95 (commented):
Code looks good just add assignment operator and it will be perfect

ziobron:
> I've created some test which will not compile since copy constructor and operator= are deleted in my_unique_ptr, i do not know how to test situation when compilation should fail

In fact you cannot test it. You can test that compilation fails when there is a copying, but it's really not so important. Compiler will protect you writing not compiling code.

homework/unique_ptr/test.cpp
```diff
@@ -0,0 +1,106 @@
...
+    EXPECT_EQ(pointer_.get(), nullptr);
+    EXPECT_EQ(*raw_pointer, 42);
+
+    delete raw_pointer;
+}
+
+TEST(Test, ShouldReset)
+{
+    int initial_value = 42;
+    int another_initial_value = 43;
```
ziobron, line 71: you can create a test suite that will hold common things. You use such values across many tests.

homework/unique_ptr/test.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+    my_unique_ptr<int> pointer_ = my_unique_ptr<int>(new int(initial_value));
+    pointer_.reset(new int(another_initial_value));
```
ziobron, line 74: test with empty reset would also be useful.

ziobron (changes requested):
A few improvements are desired :)
Please also fix @MaciejSalamonski findings.

homework/unique_ptr/test.cpp
```diff
@@ -20,37 +20,47 @@ my_unique_ptr<SomeUsefulClass> provider(std::string someString)
...
 }
 
-TEST(Test, passedWithMove)
+TEST_F(UniquePointerTestSuite, passedWithMove)
 {
-    std::string someString("provided by provider");
-    my_unique_ptr<SomeUsefulClass> ptr = provider(someString);
-    EXPECT_EQ(functionUptrAsArg(std::move(ptr)), someString);
+    my_unique_ptr<SomeUsefulClass> ptr = provider(m_someString);
+    EXPECT_EQ(functionUptrAsArg(std::move(ptr)), m_someString);
```
ziobron: please also add a test for move assignment. It will be called when you have already initialized unique_ptr and you move other initialized ptr into it.

ziobron (changes requested):
One missing test case.

homework/unique_ptr/test.cpp
```diff
@@ -59,6 +54,20 @@ TEST_F(UniquePointerTestSuite, passedWithMove)
...
+    EXPECT_EQ(m_sut.get(), nullptr);
+    EXPECT_EQ(ptr2->message_,m_msg);
+}
+
+TEST_F(UniquePointerTestSuite, moveOperatorTestWithCast)
+{
+    my_unique_ptr<SomeUsefulClass> ptr2 = std::move(m_sut);
+    EXPECT_EQ(m_sut.get(), nullptr);
+    EXPECT_EQ(ptr2->message_,m_msg);
+}
```
ziobron, line 69: duplicated test?

homework/unique_ptr/test.cpp
```diff
@@ -59,6 +54,20 @@ TEST_F(UniquePointerTestSuite, passedWithMove)
     EXPECT_EQ(functionUptrAsArg(std::move(ptr)), m_someString);
 }
 
+TEST_F(UniquePointerTestSuite, moveOperator)
+{
+    my_unique_ptr<SomeUsefulClass> ptr2 = std::move(m_sut);
```
ziobron, line 59: this calls move constructor, not assignment operator. Assignment takes place when the object was already managing another memory. Do it in this way:
```suggestion
    my_unique_ptr<SomeUsefulClass> ptr2{new SomeUsefulClass{}};
    ptr2 = std::move(m_sut);
```

ziobron (changes requested):

homework/unique_ptr/my_unique_ptr.hpp
```diff
@@ -1,40 +1,45 @@
...
 
-  T &operator=(my_unique_ptr<T> &&ptr_m) {
-    this->ptr_ = ptr_.release();
-    return *(this->ptr_);
-  }
+    T& operator=(my_unique_ptr<T>&& ptr_m)
+    {
+        this->ptr_ = ptr_m.release();
+        return *(this->ptr_);
+    }
```
ziobron, line 18: Valgrind claims that there are memory leaks in this move operator
https://github.com/marcindus/memory-management/runs/1269365447?check_suite_focus=true

You should delete a memory of the previously own object before assigning new value
And it should not return `T&`, but `my_unique_ptr<T>&`

```suggestion
    my_unique_ptr<T>& operator=(my_unique_ptr<T>&& ptr_m) noexcept
    {
        delete ptr_;
        ptr_ = ptr_m.release();
        return *this;
    }
```

homework/unique_ptr/my_unique_ptr.hpp
```diff
@@ -1,40 +1,45 @@
...
 
-  my_unique_ptr(const my_unique_ptr<T> &) = delete;
-  void operator=(const my_unique_ptr<T> &) = delete;
+    my_unique_ptr(const my_unique_ptr<T>&) = delete;
+    void operator=(const my_unique_ptr<T>&) = delete;
```
ziobron, line 21:
```suggestion
    my_unique_ptr<T>& operator=(const my_unique_ptr<T>&) = delete;
```

ziobron (changes requested):

ziobron (approved):
Good, no memory leaks :)

ziobron:
{ "requirement_ids": [3784] }

PR #42, opened 2020-10-11
Smart Pointers Homework

WojtekMs:
- shared_ptr
- weak_ptr
- make_shared


MaciejSalamonski (approved):
Everything looks good :p

homework/shared_ptr/shared.hpp
```diff
@@ -0,0 +1,184 @@
...
+    shared_ptr(const shared_ptr&) noexcept;
+    shared_ptr(shared_ptr&&) noexcept;
+    explicit shared_ptr(const weak_ptr<T>&);
+    ~shared_ptr() noexcept;
+
+    shared_ptr& operator=(shared_ptr&&) noexcept;
+    shared_ptr& operator=(const shared_ptr&) noexcept;
+
+    T* get() const noexcept { return data_; }
+    T& operator*() noexcept { return *data_; }
```
J-Cieplinski, line 49: dereferencing a nullptr

homework/shared_ptr/shared.hpp
```diff
@@ -0,0 +1,184 @@
...
+    if (this == &rhs) {
+        return *this;
+    }
+    releaseMemory();
+    data_ = rhs.data_;
+    controlBlock_ = rhs.controlBlock_;
+    if (controlBlock_) {
+        controlBlock_->incrementSharedRef();
+    }
+    return *this;
```
J-Cieplinski, line 118:
```suggestion
    if (this != &rhs) {
        releaseMemory();
         data_ = rhs.data_;
         controlBlock_ = rhs.controlBlock_;
         if (controlBlock_) {
             controlBlock_->incrementSharedRef();
         }
    }
    return *this;
```

homework/shared_ptr/shared.hpp
```diff
@@ -0,0 +1,184 @@
...
+}
+
+template <typename T>
+shared_ptr<T>& shared_ptr<T>::operator=(shared_ptr&& rhs) noexcept
```
J-Cieplinski, line 122: same here

homework/shared_ptr/shared.hpp
```diff
@@ -0,0 +1,184 @@
...
+            delete controlBlock_;
+            controlBlock_ = nullptr;
+        }
+    } else {
+        if (controlBlock_->getSharedRef() == 0) {
+            controlBlock_->getDeleter()(data_);
+            data_ = nullptr;
+            if (controlBlock_->getWeakRef() == 0) {
+                delete controlBlock_;
+                controlBlock_ = nullptr;
```
J-Cieplinski, line 159: So this first `if` is only called when there is an object. Then you check if there are any shared_ptr and delete whole control block along with that object. Only then if there is no object you check if there are any shared_ptr pointing at it and delete that object(one that  already doesn't exist)?

homework/shared_ptr/shared.hpp
```diff
@@ -0,0 +1,184 @@
...
+}
+
+template <typename T>
+void shared_ptr<T>::releaseMemory() noexcept
+{
+    if (!controlBlock_) {
+        return;
+    }
+    controlBlock_->decrementSharedRef();
+    if (controlBlock_->getObject()) {
```
J-Cieplinski, line 148: If `data_` and `controlBlock_->getObject()` are holding the same thing, why not just `if(data_)`?

homework/shared_ptr/weak.hpp
```diff
@@ -0,0 +1,138 @@
...
+        controlBlock_->decrementWeakRef();
+        if (controlBlock_->getSharedRef() == 0 && controlBlock_->getWeakRef() == 0) {
+            delete controlBlock_;
+            controlBlock_ = nullptr;
+        }
+    }
+}
+
+template <typename T>
+weak_ptr<T>& weak_ptr<T>::operator=(const weak_ptr<T>& rhs) noexcept {
```
J-Cieplinski, line 74: Same thing as with shared_ptr

homework/shared_ptr/weak.hpp
```diff
@@ -0,0 +1,138 @@
...
+    }
+    releaseMemory();
+    data_ = rhs.data_;
+    controlBlock_ = rhs.controlBlock_;
+    controlBlock_->incrementWeakRef();
+    return *this;
+}
+
+template <typename T>
+weak_ptr<T>& weak_ptr<T>::operator=(weak_ptr<T>&& rhs) noexcept
```
J-Cieplinski, line 86: as above

homework/shared_ptr/controlObject.hpp
```diff
@@ -0,0 +1,29 @@
...
+template <typename T>
+controlObject<T>::controlObject(T data, Deleter<T> deleter)
+:control_block<T>(deleter), data_(data) {
+
+}
+
+// template <typename T>
+// controlObject<T>::~controlObject() {
+//     control_block<T>::getDeleter()(data_);
+// }
```
J-Cieplinski, line 23: Comments

WBigaj (commented):
I have checked your code and have nothing to add that was not mentioned by previous reviewers.

homework/shared_ptr/shared.hpp
WojtekMs: No. This call `controlBlock_->getObject()` is in fact checking if shared_ptr was created by make_shared. In case it was created by make_shared, controlBlock will own the object and return it's address (not nullptr), in that case all I have to do is delete controlBlock [because it owns object it will delete object on it's own].
In case `controlBlock->getObject()` returns nullptr it means that my shared_ptr was not created by make_shared and controlBlock does not own an object. Due to this fact I have to delete both the object (data_) and controlBlock. But I can delete data_ only when there are no other shared_ptrs pointing to it. Similarly I can delete controlBlock only when there are no other weak_ptrs pointing to it

homework/shared_ptr/shared.hpp
WojtekMs: This is not obligatory. I prefer to write code in functions this way:
1) validate arguments
2) perform action
It's true that I have 2 return statements but in a short function like this it is not a problem, and keeping to this convention helps me to remember about validating arguments every time. Additionally my code is not nested inside if statement

homework/shared_ptr/shared.hpp
WojtekMs: `data_` and `controlBlock_->getObject()` are holding the same thing only when shared_ptr was created by make_shared. In case it was created by it's constructor then `controlBlock_->getObject()` would return nullptr whereas `data_` would hold the address of the object.

homework/shared_ptr/controlObject.hpp
WojtekMs: Thanks, I missed it

homework/shared_ptr/shared.hpp
WojtekMs: According to: [cppreference](https://en.cppreference.com/w/cpp/memory/shared_ptr/operator*) it is undefined behavior if the stored pointer is null. I'm implementing std::shared_ptr, that's why I will leave it like that.

homework/shared_ptr/control.hpp
```diff
@@ -0,0 +1,81 @@
...
+    int getSharedRef() const noexcept;
+    int getWeakRef() const noexcept;
+    void incrementSharedRef() noexcept;
+    void decrementSharedRef() noexcept;
+    void incrementWeakRef() noexcept;
+    void decrementWeakRef() noexcept;
+    virtual T* getObject() noexcept;
+    Deleter<T> getDeleter() noexcept;
+
+};
```
jakjan95, line 31: i would rather use here scissors rule(first public then private)

jakjan95 (approved):
Code looks good, just little comment from me

homework/shared_ptr/control.hpp
WojtekMs: Okay, thank you for your suggestion, but what is the benefit of this approach?

homework/shared_ptr/tests/shared_tests.cpp
```diff
@@ -0,0 +1,171 @@
...
+    auto customDeleter = [](int* data) { delete [] data; };
+    defaultShared.reset(new int[defaultValue]{}, customDeleter);
+}
+
+TEST_F(SharedTest, sharedPtrCanBeCopiedFromEmptySharedPtr) {
+    cs::shared_ptr<int> emptyShared1{};
+    cs::shared_ptr<int> emptyShared2{};
+    emptyShared1 = emptyShared2;
+    EXPECT_EQ(emptyShared1.get(), nullptr);
+}
```
ziobron, line 171: There is no need to put `shared` in every test name

homework/shared_ptr/tests/shared_tests.cpp
```diff
@@ -0,0 +1,171 @@
...
+    EXPECT_EQ(*test, anotherValue);
+}
+
+TEST_F(SharedTest, sharedArrowOperatorShoulCallFunctionsOfAssignedObject)
+{
+    struct foo {
+        void use(){};
+    };
+    cs::shared_ptr<foo> test(new foo());
+    test->use();
```
ziobron, line 106: there are no assertions so generally everything can pass this test. Checking value returned from use could do the job.

homework/shared_ptr/tests/shared_tests.cpp
```diff
@@ -0,0 +1,171 @@
...
+    test.reset();
+    EXPECT_EQ(test.get(), nullptr);
+}
+
+TEST_F(SharedTest, sharedOperatorBoolShouldCheckIfManagedObjectIsNotNullptr)
+{
+    cs::shared_ptr<int> test(new int{defaultValue});
+    EXPECT_EQ(test, true);
+    test.reset();
+    EXPECT_EQ(test, false);
```
ziobron, line 130: or just EXPECT_TRUE and EXPECT_FALSE, which are the preferred way.

homework/shared_ptr/tests/shared_tests.cpp
```diff
@@ -0,0 +1,171 @@
...
+}
+
+TEST_F(SharedTest, sharedCanBeCreatedWithCustomDeleter) {
+    int arraySize{10};
+    auto arrayDeleter = [](int* data) { delete [] data; };
+    cs::shared_ptr<int> test(new int[arraySize]{}, arrayDeleter);
+}
```
ziobron, line 137: Big 👍  for this test. Many people are not testing custom deleters.
However your test does not check it this deleter was in fact called. Of course valgrind can tell you if there are memory leaks, but if your deleter can change capture local bool value and change it, you could detect that it was indeed called.
You can do the same in resetCanBeUseWithCustomDeleter

homework/shared_ptr/tests/shared_tests.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+TEST_F(SharedTest, customDeleterShouldNotChangeSize) {
+    auto arrayDeleter = [](int* data) { delete [] data; };
+    cs::shared_ptr<int> customShared(new int[defaultValue]{}, arrayDeleter);
+    EXPECT_EQ(sizeof(customShared), sizeof(defaultShared));
+}
```
ziobron, line 143: great test!

homework/shared_ptr/tests/shared_tests.cpp
```diff
@@ -0,0 +1,171 @@
...
+TEST_F(SharedTest, resetCanBeUseWithCustomDeleter) {
+    auto customDeleter = [](int* data) { delete [] data; };
+    defaultShared.reset(new int[defaultValue]{}, customDeleter);
+}
+
+TEST_F(SharedTest, sharedPtrCanBeCopiedFromEmptySharedPtr) {
+    cs::shared_ptr<int> emptyShared1{};
+    cs::shared_ptr<int> emptyShared2{};
+    emptyShared1 = emptyShared2;
+    EXPECT_EQ(emptyShared1.get(), nullptr);
```
ziobron, line 170: you can also check use_count before and after copy

homework/shared_ptr/weak.hpp
```diff
@@ -0,0 +1,145 @@
...
+template <typename T>
+int weak_ptr<T>::use_count() const noexcept {
+    if (controlBlock_) {
+        return controlBlock_->getSharedRef();
+    }
+    return 0;
+}
+
+template <typename T>
+
```
ziobron, line 130:
```suggestion
```

homework/shared_ptr/tests/weak_tests.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+};
+
+TEST_F(WeakTest, canBeDefaultCreated) {
+    cs::weak_ptr<int> ptr{};
+}
+
+TEST_F(WeakTest, canBeCreatedFromSharedPtr) {
+    cs::weak_ptr<int> ptr{defaultShared};
+}
```
ziobron, line 20: please add some assertions, eg. on `expired()` (or merge it some futher tests)

homework/shared_ptr/tests/weak_tests.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+TEST_F(WeakTest, expiredShouldBeFalseWhenNotDangling) {
+    cs::weak_ptr<int> ptr{defaultShared};
+    EXPECT_EQ(ptr.expired(), false);
```
ziobron, line 24: you may use EXPECT_FALSE / TRUE here and somewhere below

homework/shared_ptr/tests/weak_tests.cpp
```diff
@@ -0,0 +1,132 @@
...
+    defaultShared.reset();
+    EXPECT_EQ(ptr.expired(), true);
+}
+
+TEST_F(WeakTest, useCountShouldReturnTheNumberOfSharedReferences) {
+    cs::weak_ptr<int> ptr{defaultShared};
+    int expectedUseCount{1};
+    EXPECT_EQ(ptr.use_count(), defaultShared.use_count());
+    EXPECT_EQ(ptr.use_count(), expectedUseCount);
+    EXPECT_EQ(defaultShared.use_count(), expectedUseCount);
```
ziobron, line 38: you can copy a weak_ptr here and check if use_count has not incremented, but it is incremented when a copy of shared_ptr is made.

homework/shared_ptr/tests/weak_tests.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+TEST_F(WeakTest, lockShouldReturnNullptrWhenNoManagedObject) {
+    cs::weak_ptr<int> ptr{defaultShared};
+    ptr.reset();
+    EXPECT_EQ(ptr.lock().get(), nullptr);
+}
+
+TEST_F(WeakTest, lockShouldReturnSharedPtrWhenIsManagedObject) {
+    cs::weak_ptr<int> ptr{defaultShared};
+    EXPECT_EQ(ptr.lock(), defaultShared);
```
ziobron, line 67: do we have comparison operator on shared_ptrs? Does it work as expected? I guess I have not seen your implementation. You can of course compare dereferenced objects.

homework/shared_ptr/tests/weak_tests.cpp
```diff
@@ -0,0 +1,132 @@
...
+}
+
+TEST_F(WeakTest, lockShouldReturnNullPtrAfterWeakPtrReset) {
+    cs::weak_ptr<int> ptr{defaultShared};
+    ptr.reset();
+    EXPECT_EQ(ptr.lock().get(), nullptr);
+}
```
ziobron, line 74: copy of lockShouldReturnNullptrWhenNoManagedObject (line 59)

homework/shared_ptr/tests/weak_tests.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+TEST_F(WeakTest, copyAssignmentShouldIncreaseWeakRefCount) {
```
ziobron, line 76: this name is not corresponding with what this test does check. Neither a few below tests where you mention WeakRefCount. It's in control block and we have no way to check it here.

ziobron (changes requested):
Nice implementation. You have really good tests on ref counting 👍 
Thanks @J-Cieplinski for insightful review and @WojtekMs for accurate answers 👍
I was able to find 1 major issue in make_shared.

homework/shared_ptr/bad_weak_ptr.hpp
```diff
@@ -0,0 +1,14 @@
...
+
+namespace cs {
+    class bad_weak_ptr :public std::exception {
+        std::string message{"weak ptr is expired!\n"};
+        public:
+        bad_weak_ptr() = default;
+        explicit bad_weak_ptr(const std::string& what) :message(what) {}
+        const char* what() const noexcept override {
+            return message.c_str();
+        }
```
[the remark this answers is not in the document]
WojtekMs: However, if I inherit from std::exception and I do not override `what()` method, it is printing std::exception not my own cs::bad_weak_ptr type

homework/shared_ptr/shared.hpp
```diff
@@ -0,0 +1,184 @@
...
+    friend class weak_ptr;
+
+    template <typename U, typename... Ts>
+    friend shared_ptr<U> make_shared(Ts&&...);
+
+    T* data_{nullptr};
+    control_block<T>* controlBlock_{nullptr};
+
+    void releaseMemory() noexcept;
+    shared_ptr(control_block<T>*) noexcept;
```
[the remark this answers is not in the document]
WojtekMs: In my implementation of make_shared
```cpp
template <typename T, typename... Ts>
shared_ptr<T> make_shared(Ts&&... args)
{
    return shared_ptr<T>{new controlObject<T, Ts...>(std::forward<Ts>(args)...)};
}
I create shared_ptr out of controlObject, which is derived from control_block.
```

homework/shared_ptr/tests/weak_tests.cpp
WojtekMs: Good point, the test lockShouldReturnNullptrWhenNoManagedObject was supposed to reset shared_ptr not weak_ptr

homework/shared_ptr/tests/weak_tests.cpp
WojtekMs: You are right. I corrected this test name, but when it comes to checking if memory is released when we use copy / move operators it can be only checked by valgrind and ie. weakWhenCopiedShouldReleaseMemoryWhenIsTheLastWeakPtrAndSharedCountIs0 test scenario is created for this purpose.

WojtekMs:
As requested:
- make_shared fix + tests checking many arguments are added
- bad_weak_ptr what override removed [but now what method returns `std::exception` instead of `std::bad_weak_ptr`
- comparing of 2 shared_ptr's is corrected - now dereferenced shared_ptrs are compared
- misleading test name corrected

homework/shared_ptr/tests/weak_tests.cpp
WojtekMs: There is no operator== implemented for `cs::shared_ptr`. I guess they must've been converted to boolean values and then compared

homework/shared_ptr/bad_weak_ptr.hpp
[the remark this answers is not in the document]
WojtekMs: However, if I remove my implementation `what()` returns `std::exception` instead of `std::bad_weak_ptr`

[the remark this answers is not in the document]
ziobron: how/where do you catch it?

homework/shared_ptr/tests/shared_tests.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+class SharedTest : public ::testing::Test
+{
+   protected:
+    int defaultValue{5};
+    int anotherValue{10};
+    cs::shared_ptr<int> defaultShared{new int{defaultValue}};
+    void release() {
+    throw std::exception();
+}
```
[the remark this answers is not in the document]
ziobron: Maybe you get std::exception on the output because of this?

ziobron (approved):
Good job 👍

ziobron:
{ "requirement_ids": [3794, 3795, 3796] }

homework/shared_ptr/tests/shared_tests.cpp
WojtekMs: That's true :D

ziobron:
{
"requirement_ids": [1119, 1119, 1119]
}

PR #44, opened 2020-10-13
shared_ptr/make_shared/weak_ptr

J-Cieplinski:
@J-Cieplinski
@majkel84


MaciejSalamonski (approved):
Very nice code ;~} 👍

WBigaj (approved):
I have checked your code and it looks great 👍

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,180 @@
...
+template <typename T>
+class weak_ptr;
+
+template <typename T>
+class shared_ptr {
+public:
+    template <typename B> friend class weak_ptr;
+    template<typename D, typename... Args> friend shared_ptr<D> make_shared(Args&& ... args);
+    explicit shared_ptr() = default;
+    explicit shared_ptr(T* ptr);
```
WojtekMs, line 18: I think you do not need explicit in the default c-tor ;)

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,180 @@
...
+    shared_ptr(shared_ptr&& previousOwner) noexcept;      //move c-tor
+    ~shared_ptr();
+
+    void swap(shared_ptr<T>& secondPointer) noexcept;
+    const T* get() const;
+    void reset(T* newPtr = nullptr, std::function<void(T*)> deleter = [](T* ptr) { delete ptr;});
+
+    const T* operator->();
+    T& operator*();
+    explicit operator bool() const noexcept;
```
WojtekMs, line 31: I think operator bool() should not be explicit. This way in if statement you would have to write code like:
`if(bool(ptr))`

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,180 @@
...
+}
+
+template <typename T>
+const T* shared_ptr<T>::get() const {
+    return ptr_;
+}
+
+template <typename T>
+void shared_ptr<T>::reset(T* newPtr, std::function<void(T*)> deleter) {
+    shared_ptr{newPtr, deleter}.swap(*this);
```
WojtekMs, line 122: Nice one :)

homework/shared_ptr/control_block.hpp
```diff
@@ -0,0 +1,66 @@
...
+    
+};
+
+template <typename T>
+class continuous_block : public control_block<T> {
+public: 
+    
+    template<typename ...Args>
+    continuous_block(Args&& ...args) : control_block<T>([](T*) {}), object_{std::forward<Args>(args)...} {
+    }
```
WojtekMs, line 55: Why are you passing an empty lambda as a deleter?

homework/shared_ptr/weak_ptr.hpp
```diff
@@ -0,0 +1,125 @@
...
+#include "control_block.hpp"
+#include "shared_ptr.hpp"
+
+namespace cs {
+template <typename T>
+class weak_ptr {
+public:
+    constexpr weak_ptr() noexcept = default;
+    weak_ptr(const weak_ptr& ptr) noexcept;
+    weak_ptr(const shared_ptr<T>& ptr) noexcept;
```
WojtekMs, line 13: I think it would be good to add explicit to c-tor with shared_ptr

homework/shared_ptr/shared_ptr.hpp
J-Cieplinski: we don't have to

homework/shared_ptr/weak_ptr.hpp
```diff
@@ -0,0 +1,125 @@
...
+}
+
+template <typename T>
+weak_ptr<T>& weak_ptr<T>::operator=(const weak_ptr& ptr) noexcept {
+    ptr_ = ptr.ptr_;
+    counter_ = ptr.counter_;
+    if (counter_) {
+        counter_->increaseWeakRefs();
+    }
+    return *this;
```
WojtekMs, line 80: remember to check if you are not self-assigning and to decreaseWeakRefs before `counter_ = ptr.counter_`

homework/shared_ptr/weak_ptr.hpp
```diff
@@ -0,0 +1,125 @@
...
+}
+
+template <typename T>
+weak_ptr<T>& weak_ptr<T>::operator=(const shared_ptr<T>& ptr) noexcept {
+    ptr_ = ptr.ptr_;
+    counter_ = ptr.counter_;
+    if (counter_) {
+        counter_->increaseWeakRefs();
+    }
+    return *this;
```
WojtekMs, line 90: remember to decrease weakRefs :)

homework/shared_ptr/weak_ptr.hpp
```diff
@@ -0,0 +1,125 @@
...
+}
+
+
+template <typename T>
+weak_ptr<T>& weak_ptr<T>::operator=(weak_ptr&& previousOwner) noexcept {
+    if (this != &previousOwner) {
+        weak_ptr(previousOwner).swap(*this);
+        previousOwner.ptr_ = nullptr;
+        previousOwner.counter_ = nullptr;
+    }
```
WojtekMs, line 100: This code works, is very short and very smart :) However, move operations should be treated as an optimization. I think creating temporary object is not the best way to do it.

homework/shared_ptr/weak_ptr.hpp
```diff
@@ -0,0 +1,125 @@
...
+}
+
+template <typename T>
+void weak_ptr<T>::reset() noexcept {
+    weak_ptr{}.swap(*this);
+}
+
+template <typename T>
+bool weak_ptr<T>::expired() const noexcept {
+    return (counter_ == nullptr || counter_->getRefs() == 0);
```
WojtekMs, line 111: smart :)

WojtekMs (commented):
Good code :)
shared_ptr(const weak_ptr<T>&) would come in handy though

homework/shared_ptr/control_block.hpp
J-Cieplinski: Nature of continuous_block. I don't want to delete T if sharedRefs get to 0. Nature of `if any std::weak_ptr references the control block created by std::make_shared after the lifetime of all shared owners ended, the memory occupied by T persists until all weak owners get destroyed as well,`

homework/shared_ptr/shared_ptr.hpp
WojtekMs: True, I didn't know about contextual conversions. It's interesting, thanks

jakjan95 (approved):
For me everything looks good +1

homework/shared_ptr/weak_ptr.hpp
majkel84: done

homework/shared_ptr/weak_ptr.hpp
majkel84: done

homework/shared_ptr/weak_ptr.hpp
majkel84: done

homework/shared_ptr/control_block.hpp
```diff
@@ -0,0 +1,66 @@
...
+template <typename T>
+class control_block {
+public:
+    control_block(std::function<void(T*)> deleter = [](T* ptr) { delete ptr;} )
+        : deleter_(deleter), sharedRefs_(0), weakRefs_(0) {};
+    control_block(const control_block&) = delete; 
+    control_block& operator=(const control_block&) = delete;
+    virtual ~control_block() {
+        sharedRefs_ = 0;
+        weakRefs_ = 0;
```
ziobron, line 16: it's not needed. Zeroing memory takes additional time on destruction.

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,180 @@
...
+class weak_ptr;
+
+template <typename T>
+class shared_ptr {
+public:
+    template <typename B> friend class weak_ptr;
+    template<typename D, typename... Args> friend shared_ptr<D> make_shared(Args&& ... args);
+    explicit shared_ptr() = default;
+    explicit shared_ptr(T* ptr);
+    shared_ptr(T* ptr, std::function<void(T*)> deleter);
```
ziobron, line 19: you can merge this constructor with above by adding a default parameter value to deleter

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,180 @@
...
+    shared_ptr(const shared_ptr& ptr) noexcept;  //copy c-tor
+    shared_ptr(shared_ptr&& previousOwner) noexcept;      //move c-tor
+    ~shared_ptr();
+
+    void swap(shared_ptr<T>& secondPointer) noexcept;
+    const T* get() const;
+    void reset(T* newPtr = nullptr, std::function<void(T*)> deleter = [](T* ptr) { delete ptr;});
+
+    const T* operator->();
+    T& operator*();
```
ziobron, line 30:
```suggestion
    const T* operator->() const;
    T& operator*() const;
```

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,180 @@
...
+    explicit operator bool() const noexcept;
+    shared_ptr<T>& operator=(const shared_ptr<T>& ptr) noexcept;             //copy assignment
+    shared_ptr<T>& operator=(shared_ptr<T>&& previousOwner) noexcept;  //move assignment
+
+    size_t use_count() { return counter_->getRefs(); }
```
ziobron, line 35:
```suggestion
    size_t use_count() const { return counter_->getRefs(); }
```

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,180 @@
...
+    void checkAndDeletePointers();
+};
+
+template <typename T>
+void shared_ptr<T>::checkAndDeletePointers() {
+    if (!counter_->getRefs()) {
+            counter_->deleter_(ptr_);
+        if (!counter_->getWeakRefs()) {
+            delete counter_;
+        }
```
ziobron, line 51:
```suggestion
    if (!counter_->getRefs()) {
        counter_->deleter_(ptr_);
        if (!counter_->getWeakRefs()) {
            delete counter_;
        }
```

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,180 @@
...
+    shared_ptr(T* ptr, control_block<T>* counter);
+    T* ptr_{nullptr};
+    control_block<T>* counter_{nullptr};
+
+    void checkAndDeletePointers();
+};
+
+template <typename T>
+void shared_ptr<T>::checkAndDeletePointers() {
+    if (!counter_->getRefs()) {
```
ziobron, line 47: you may add decrementing here before any checks.

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,180 @@
...
+            counter_->deleter_(ptr_);
+        if (!counter_->getWeakRefs()) {
+            delete counter_;
+        }
+    }
+}
+
+template <typename T>
+shared_ptr<T>::shared_ptr(T* ptr, control_block<T>* counter) : ptr_(ptr), counter_(counter) {
+    ++*counter_;
```
ziobron, line 57: your control block may be created by default with 1. There will be no need to increment it here and in some other places.

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,180 @@
...
+}
+
+template <typename T>
+void shared_ptr<T>::swap(shared_ptr<T>& secondPointer) noexcept {
+    auto ptrTmp = secondPointer.ptr_;
+    auto counterTmp = secondPointer.counter_;
+    secondPointer.ptr_ = ptr_;
+    secondPointer.counter_ = counter_;
+    ptr_ = ptrTmp;
+    counter_ = counterTmp;
```
ziobron, line 112: you may call std::swap 2 times instead of those 6 lines.

homework/shared_ptr/shared_ptr.hpp
ziobron: it may be unsafe. You need a test as a proof that it works correctly. What about such case:
```cpp
cs::shared_ptr<int> sh{};
int* p = new int{5};
sh.reset(p);
p = 10;
```

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,180 @@
...
+
+template <typename T>
+const T* shared_ptr<T>::operator->() {
+    return ptr_;
+}
+
+template <typename T>
+T& shared_ptr<T>::operator*() {
+    if(!ptr_) {
+        throw std::runtime_error("Dereferencing a nullptr");
```
ziobron, line 133: it's not a normal shared_ptr behavior. I guess that you wanted to have your own behavior here.

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,180 @@
...
+        previousOwner.ptr_ = nullptr;
+        previousOwner.counter_ = nullptr;
+    }
+
+    return *this;
+}
+
+template <typename T>
+shared_ptr<T>& shared_ptr<T>::operator=(const shared_ptr<T>& ptr) noexcept {
+    if(counter_){
```
ziobron, line 162: what about self assignment here?

homework/shared_ptr/test.cpp
```diff
@@ -0,0 +1,222 @@
...
+    auto sPtr3 = sPtr;
+    ASSERT_EQ(*sPtr, *sPtr2);
+    ASSERT_EQ(*sPtr, *sPtr3);
+    ASSERT_EQ(sPtr3.use_count(), 3);
+}
+
+TEST_F(sharedPtrTest, testCopyAssignment) {
+    ASSERT_EQ(*sPtr, testValueOne);
+    ASSERT_EQ(sPtr.use_count(), 1);
+    cs::shared_ptr<int> sPtr2;
```
ziobron, line 48:
```suggestion
    cs::shared_ptr<int> sPtr2{new int{100}};
```
This may cause some memory leaks if you have bad implementation, so it is better to use it in tests. The same in other places and tests like testMoveAssignment.

homework/shared_ptr/test.cpp
```diff
@@ -0,0 +1,222 @@
...
+    ASSERT_EQ(*sPtr, testValueTwo);
+    ASSERT_EQ(*testPtr, testValueOne);
+    ASSERT_EQ(testPtr.use_count(), higherUseCount);
+    ASSERT_EQ(sPtr.use_count(), lowerUseCount);
+}
+
+TEST_F(sharedPtrTest, testAccessOperator) {
+    cs::shared_ptr<std::string> uPtr2(new std::string{testString});
+    ASSERT_EQ(uPtr2->at(0), 'A');
+    ASSERT_EQ(uPtr2->at(1), 'l');
```
ziobron, line 100:
```suggestion
    ASSERT_EQ(uPtr2->at(0), testString[0]);
    ASSERT_EQ(uPtr2->at(1), testString[1]);
```
Write code in such way that will require you less changes in future. With my bersion it will work even if you change the content of testString.

homework/shared_ptr/test.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+    {
+        auto ptr = cs::make_shared<A>(testNumber, testString);
+        wPtr = ptr;
+        object = ptr.get();
+    }
+
+    ASSERT_EQ(testNumber, object->a);
+    ASSERT_EQ(testString, object->b);
+}
```
ziobron, line 164: very good tests!

ziobron (changes requested):
exercises and unique_ptr should not be commited. I have checked only shared_ptr and make_shared part. Let me check weak_ptr later.

homework/shared_ptr/shared_ptr.hpp
J-Cieplinski: I tried, it would not work. If I recall correctly it was first checking for matching function, and in that case both `shared_ptr()`(which was deleted) and overloaded `T* ptr = nullptr` were detected and there was compilation error. I've read that first the compiler tries to use correct overloaded function and then checks for deleted functions etc.

homework/shared_ptr/test.cpp
```diff
@@ -0,0 +1,222 @@
...
+TEST_F(weakPtrTest, testLock) {
+    cs::weak_ptr<int> wPtr(sPtr);
+    ASSERT_EQ(*(wPtr.lock()), *sPtr);
+}
+
+TEST_F(weakPtrTest, testCopyConstructorFromShared) {
+    cs::weak_ptr<int> wPtr(sPtr);
+    ASSERT_EQ(*(wPtr.lock()), *sPtr);
+    ASSERT_EQ(wPtr.use_count(), 1);
+}
```
ziobron, line 182: in fact you can merge those 2 tests. Their bodies are almost the same.

homework/shared_ptr/test.cpp
```diff
@@ -0,0 +1,222 @@
...
+TEST_F(weakPtrTest, testCopyConstructorFromWeak) {
+    cs::weak_ptr<int> wPtr(sPtr);
+    auto wPtr2(wPtr);
+    ASSERT_FALSE(wPtr.expired());
+    ASSERT_EQ(*(wPtr.lock()), *sPtr);
+    ASSERT_EQ(*(wPtr2.lock()), *sPtr);
+}
+
+TEST_F(weakPtrTest, testCopyAssignment) {
+    cs::weak_ptr<int> wPtr = sPtr;
```
ziobron, line 193: it's not copy assignment. It will take place if your wPtr was already pointing to something else.

homework/shared_ptr/test.cpp
```diff
@@ -0,0 +1,222 @@
...
+    cs::weak_ptr<int> wPtr{sPtr};
+    auto wPtr2(std::move(wPtr));
+    ASSERT_EQ(*(wPtr2.lock()), *sPtr);
+    ASSERT_EQ(wPtr2.use_count(), 1);
+    ASSERT_TRUE(wPtr.expired());
+}
+
+TEST_F(weakPtrTest, testMoveAssignment) {
+    cs::weak_ptr<int> wPtr{sPtr};
+    auto wPtr2 = std::move(wPtr);
```
ziobron, line 208: it's move constructor. The same as for copy - wPtr2 needs to be initialized before assignment to call move assignment operator.

homework/shared_ptr/weak_ptr.hpp
```diff
@@ -0,0 +1,132 @@
...
+weak_ptr<T>& weak_ptr<T>::operator=(const weak_ptr& ptr) noexcept {
+    if (this != &ptr) {
+        if(counter_) {
+            counter_->decreaseWeakRefs();
+            checkAndDeletePointers();
+        }
+    ptr_ = ptr.ptr_;
+    counter_ = ptr.counter_;
+        counter_->increaseWeakRefs();
+    }
```
ziobron, line 83:
```suggestion
    if (this != &ptr) {
        if(counter_) {
            counter_->decreaseWeakRefs();
            checkAndDeletePointers();
        }
        ptr_ = ptr.ptr_;
        counter_ = ptr.counter_;
        counter_->increaseWeakRefs();
    }
```

homework/shared_ptr/weak_ptr.hpp
```diff
@@ -0,0 +1,132 @@
...
+    return *this;
+}
+
+template <typename T>
+weak_ptr<T>& weak_ptr<T>::operator=(weak_ptr&& previousOwner) noexcept {
+    if (this != &previousOwner) {
+        weak_ptr(previousOwner).swap(*this);
+        previousOwner.ptr_ = nullptr;
+        previousOwner.counter_ = nullptr;
```
ziobron, line 106: I can smell a possible memory leak here. You have not tested this operator. In test move constructor is called.
Especially check it with a case when we have 2 or more weak_ptrs (pointing to the same thing) and we want to move one of them to something else using this operator. Check it please against such a scenario.

ziobron (changes requested):
Review finished.

homework/shared_ptr/test.cpp
majkel84: done

homework/shared_ptr/test.cpp
majkel84: done

homework/shared_ptr/test.cpp
majkel84: done

homework/shared_ptr/weak_ptr.hpp
majkel84: done

homework/shared_ptr/weak_ptr.hpp
majkel84: yes You right :) additional test case added.
reset() for previous owner added to operator=

homework/shared_ptr/control_block.hpp
majkel84: done

homework/shared_ptr/shared_ptr.hpp
majkel84: done

homework/shared_ptr/shared_ptr.hpp
majkel84: done

homework/shared_ptr/shared_ptr.hpp
majkel84: done

homework/shared_ptr/shared_ptr.hpp
majkel84: done

homework/shared_ptr/test.cpp
majkel84: done
we ave good implementation - no memory leaks here :)

homework/shared_ptr/test.cpp
majkel84: done

homework/shared_ptr/shared_ptr.hpp
ziobron: Yes, but do not give default value for ptr. Give it to deleter only

ziobron (approved):

ziobron:
{ "requirement_ids": [3794, 3795, 3796], "users": ["J-Cieplinski", "majkel84"] }

ziobron:
{
"requirement_ids": [1636],
"users": ["J-Cieplinski", "majkel84"]
}

PR #46, opened 2020-10-27
Homework make unique

DamiXioR:


MaciejSalamonski (approved):
I think that homework make_unique should be in repo Templates

homework/unique_ptr/UniquePointer.hpp
```diff
@@ -0,0 +1,129 @@
...
+    pointer_ = pointer;
+}
+
+template <class T>
+void UniquePointer<T>::reset()
+{
+    if (pointer_ != nullptr) {
+        delete pointer_;
+        pointer_ = nullptr;
+    }
```
WBigaj, line 108: I think this if is uncessary,because deleting nullptr is safe.

homework/unique_ptr/UniquePointer.hpp
```diff
@@ -0,0 +1,129 @@
...
+{
+}
+
+template <class T>
+UniquePointer<T>::UniquePointer(UniquePointer&& anotherUniquePointerToMove)
+{
+    if (pointer_ != nullptr) {
+        delete pointer_;
+        pointer_ = nullptr;
+    }
```
WBigaj, line 50: I think this if is uncessary,because deleting nullptr is safe.

WBigaj (commented):
I have checked your code and I think you can skip few ifs,because deleting nullptr is safe.

jakjan95 (commented):
I think that for make_unique you should make PR to templates repo :)

homework/unique_ptr/UniquePointer.hpp
```diff
@@ -117,3 +121,9 @@ T* UniquePointer<T>::release()
     pointer_ = nullptr;
     return temp;
 }
+
+template<typename MUType, typename... Args>
+UniquePointer<MUType> MakeUnique(Args&&... args)
+{
+    return UniquePointer<MUType>(new MUType(std::forward<Args>(args)...));
+}
```
ziobron, line 129: what about arrays? This version cannot handle arrays properly.

homework/unique_ptr/MakeUnique.ut.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+        WHEN("Creating UniquePointer with std::move and object created by MakeUnique")
+        {
+            UniquePointer<int> uniqueIntRValue2(std::move(uniqueIntRValue));
+            UniquePointer<int> uniqueIntLValue2(std::move(uniqueIntLValue));
+            UniquePointer<TestingClass> uniqueTestingClassRValue2(std::move(uniqueTestingClassRValue));
+            UniquePointer<TestingClass> uniqueTestingClassLValue2(std::move(uniqueTestingClassLValue));
+            UniquePointer<std::pair<std::string, int>> uniquePointerForMoreArguments2(std::move(uniquePointerForMoreArguments));
+
+            THEN("Should create unique pointer using move constructor and free memory")
```
ziobron, line 47: ok, but how to check that move constructor was called?

homework/unique_ptr/MakeUnique.ut.cpp
```diff
@@ -0,0 +1,219 @@
...
+                REQUIRE(*anyPtr2 == firstNumber);
+
+                REQUIRE(uniqueTestingClassRValue.get() == nullptr);
+                REQUIRE((*anyPtr3).testingGetter() == "any text");
+                REQUIRE(uniqueTestingClassLValue.get() == nullptr);
+                REQUIRE((*anyPtr4).testingGetter() == firstText);
+
+                REQUIRE(uniquePointerForMoreArguments.get() == nullptr);
+                REQUIRE((*uniquePointerForMoreArguments).first == "r value text");
+                REQUIRE((*uniquePointerForMoreArguments).second == firstNumber);
```
ziobron, line 95: it does not test make_unique. It is unique_ptr test. You can remove it.

ziobron (changes requested):
Generally, you test unique_ptr here, not make_unique. You have covered scenario that assures that your implementation uses variadic templates (can pass any number of parameters to make_unique).
I want to see scenarios for detecting if perfect forwarding was used, eg. if the proper move or copy constructor was called.
There should also be a scenario for creating arrays with make_unique.

homework/unique_ptr/UniquePointer.hpp
DamiXioR: i changed all implementation, now it works with std::unique_ptr, i had problems with my own UniquePointer implementation, so i cant handle arrays

DamiXioR:
i change UniquePointer implementation and now it works with arrays, make unique should works too

homework/unique_ptr/UniquePointer.hpp
```diff
@@ -122,8 +124,141 @@ T* UniquePointer<T>::release()
     return temp;
 }
 
-template<typename MUType, typename... Args>
-UniquePointer<MUType> MakeUnique(Args&&... args)
+//Array Specialization
+
+template <class T>
+class UniquePointer<T[]> {
```
ziobron, line 130: no need for such class duplication. You can only provide a proper deleter to casual unique pointer implementation. That's the only difference. Just make your delete functions replaceable.
1. Extract all delete usages into deleter function.
2. Add a uniqe_ptr constructor that also takes a deleter. Default deleter is just calling delete.
3. Use it in make_unique for arrays

homework/unique_ptr/MakeUnique.ut.cpp
```diff
@@ -15,204 +14,68 @@ SCENARIO("Using make unique to initialize unique pointers", "[makeUniquePtr]")
...
-                REQUIRE_THROWS_AS(*uniqueIntLValue, DereferenceNullPtr);
-                REQUIRE_THROWS_AS((*uniqueTestingClassRValue).testingGetter(), DereferenceNullPtr);
-                REQUIRE_THROWS_AS((*uniqueTestingClassLValue).testingGetter(), DereferenceNullPtr);
-                REQUIRE_THROWS_AS((*uniquePointerForMoreArguments).first, DereferenceNullPtr);
+                for (int i = 0; i < arraySize; i++) {
+                    REQUIRE(uniqueDoubleArray[i] == arraySize);
+                }
             }
         }
     }
```
ziobron, line 81: please add tests for lvalue and rvalue check. Proper functions/constructors should be used on move/copy operations. If you don't know what is it about take a look into other's PR in templates repo.

ziobron (changes requested):

homework/unique_ptr/UniquePointer.hpp
DamiXioR: i left it with specialization, it was to hard for me to do it with one class, i added deleters

homework/unique_ptr/UniquePointer.hpp
```diff
@@ -140,30 +156,32 @@ class UniquePointer<T[]> {
     T& operator*();
     T* get() const;
     T* release();
-    void reset();
-    void reset(T* pointer);
+    void reset(std::function<void(T*)> deleter = defaultBlockDeleter<T>);
+    void reset(T* pointer, std::function<void(T*)> deleter = defaultBlockDeleter<T>);
```
ziobron, line 160: no need to have both functions. Do sth like this:
```suggestion
    void reset(T* pointer = nullptr, std::function<void(T*)> deleter = defaultBlockDeleter<T>);
```

ziobron (approved):
I really don't like this duplication but let it be ;)

ziobron (changes requested):
Aaa and tests for rvalue and lvalue forwarding are still missing! Please add them

ziobron (approved):
👍

ziobron:
{ "requirement_ids": [4443], "points": [40] }
40 XP za Szablony #1 – make_unique (4443)

PR #47, opened 2020-10-29
Homework shared_ptr, weak_ptr, make_shared

Kuba-Wi:
Collaboration with kuchichan


MaciejSalamonski (approved):
Imo it look ok +1

WBigaj (approved):
I have checked your code and it looks good 👍

homework/shared_ptr/weak_ptr-ut.cpp
```diff
@@ -0,0 +1,68 @@
...
+TEST(WeakPointerTests, copyAssignmentOperatorShouldMovePointer) {
+    constexpr size_t numberOfUseCount = 1;
+    struct Something {};
+    cs::shared_ptr<Something> shared_something(new Something{});
+    cs::weak_ptr<Something> weak_something{shared_something};
+    cs::weak_ptr<Something> weak_ptr;
+    weak_ptr = weak_something;
+
+    ASSERT_EQ(weak_ptr.use_count(), numberOfUseCount);
+    ASSERT_EQ(weak_something.use_count(), numberOfUseCount);
```
ziobron, line 56: you may also show that before copy use_count was 0. The same in move in a test above.

homework/shared_ptr/weak_ptr-ut.cpp
```diff
@@ -0,0 +1,68 @@
...
+}
+
+TEST(WeakPointerTests, WeakPointerResetShouldSetPtrAndControlBlockAsNullPtr) {
+  int expectedValue{100};
+  auto s_ptr = cs::shared_ptr<int>(new int{20});
+  auto next_ptr = cs::shared_ptr<int>(new int{expectedValue});
+  cs::weak_ptr<int> weakSomething{s_ptr};
+  weakSomething.reset();
```
ziobron, line 64: you should also check if after reset we have use_count == 0 or pointer is expired and it is not after reassignment.

ziobron (changes requested):

homework/shared_ptr/weak_ptr.hpp
```diff
@@ -0,0 +1,123 @@
...
+    cb_ = ptr.cb_;
+    cb_->increaseWeakRef();
+}
+
+template <typename T>
+weak_ptr<T>::weak_ptr(weak_ptr&& ptr) noexcept {
+    if (this != &ptr) {
+        ptr_ = ptr.ptr_;
+        cb_ = ptr.cb_;
+        cb_->setDeleter([&]() { delete ptr_; });
```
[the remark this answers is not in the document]
Kuba-Wi: It is needed because otherwise deleter will call delete on old pointer that doesn't point on anything.

homework/shared_ptr/weak_ptr.hpp
```diff
@@ -0,0 +1,123 @@
...
+    return *this;
+}
+
+template <typename T>
+weak_ptr<T>& weak_ptr<T>::operator=(weak_ptr&& ptr) noexcept {
+    if (&ptr != this) {
+        deletePointers();
+        ptr_ = ptr.ptr_;
+        cb_ = ptr.cb_;
+        cb_->setDeleter([&]() { delete ptr_; });
```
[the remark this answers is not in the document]
Kuba-Wi: It is needed because otherwise deleter will call delete on old pointer that doesn't point on anything.

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,121 @@
...
+    ptr_ = ptr.ptr_;
+    cb_ = ptr.cb_;
+    cb_->increaseSharedRef();
+}
+
+template <typename T>
+shared_ptr<T>::shared_ptr(shared_ptr&& ptr) {
+    ptr_ = ptr.ptr_;
+    cb_ = ptr.cb_;
+    cb_->setDeleter([&]() { delete ptr_; });
```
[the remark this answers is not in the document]
Kuba-Wi: It is needed because otherwise deleter will call delete on old pointer that doesn't point on anything.

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,121 @@
...
+    return *this;
+}
+
+template <typename T>
+shared_ptr<T>& shared_ptr<T>::operator=(shared_ptr&& ptr) {
+    if (&ptr != this) {
+        deletePointers();
+        ptr_ = ptr.ptr_;
+        cb_ = ptr.cb_;
+        cb_->setDeleter([&]() { delete ptr_; });
```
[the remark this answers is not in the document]
Kuba-Wi: It is needed because otherwise deleter will call delete on old pointer that doesn't point on anything.

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,121 @@
...
+    shared_ptr(const shared_ptr& ptr);
+    shared_ptr(const cs::weak_ptr<T>& ptr);
+    shared_ptr(shared_ptr&& ptr);
+    shared_ptr& operator=(const shared_ptr& ptr);
+    shared_ptr& operator=(shared_ptr&& ptr);
+    ~shared_ptr();
+    T* get() const { return ptr_; }
+    T& operator*() const { return *ptr_; }
+    T* operator->() const { return ptr_; }
+    long use_count() const { return static_cast<long>(cb_->getShared()); }
```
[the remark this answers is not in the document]
Kuba-Wi: Declaration on cppreference:
long use_count() const noexcept;

homework/shared_ptr/control_block.hpp
```diff
@@ -3,31 +3,31 @@
 #include <atomic>
 #include <functional>
 
+template <typename T>
 class ControlBlock {
 public:
-    ControlBlock(std::function<void()> del) : deleter(del) {}
+    ControlBlock() = default;
```
ziobron, line 9: with only default constructor, you don't have a way to provide a custom deleter.

homework/shared_ptr/control_block.hpp
```diff
@@ -3,31 +3,31 @@
...
 
 template <typename T>
-class BlockAndData : public ControlBlock {
+class BlockAndData : public ControlBlock<T> {
 public:
     template <typename... Args>
-    BlockAndData<T>(Args... args) : object_(args...), ControlBlock([]() {}) {}
-    T* getObject() { return &object_; }
+    BlockAndData<T>(Args... args) : object_(args...), ControlBlock<T>() {}
+    const T& getObject() const { return object_; }
```
ziobron, line 30: why not return T*? you cannot modify underlying object with const&. Previous version was ok.

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -14,8 +14,7 @@ namespace cs {
 template <typename T>
 class shared_ptr {
 public:
-    shared_ptr(T* ptr = nullptr)
-        : ptr_(ptr), cb_(new ControlBlock([&]() { delete ptr_; })) {}
+    shared_ptr(T* ptr = nullptr) : ptr_(ptr), cb_(new ControlBlock<T>) {}
```
ziobron, line 17: passing a deleter here was really fine. Bring back the previous version :D

homework/shared_ptr/control_block.hpp
```diff
@@ -3,31 +3,31 @@
...
 class ControlBlock {
 public:
-    ControlBlock(std::function<void()> del) : deleter(del) {}
+    ControlBlock() = default;
     void increaseSharedRef() { shared_refs += 1; }
     void decreaseSharedRef() { shared_refs -= 1; }
     void increaseWeakRef() { weak_refs += 1; }
     void decreaseWeakRef() { weak_refs -= 1; }
     std::atomic<size_t>& getShared() { return shared_refs; }
     std::atomic<size_t>& getWeak() { return weak_refs; }
```
ziobron: return size_t not atomic<size_t>

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -38,17 +37,17 @@ class shared_ptr {
     template <typename>
     friend class cs::weak_ptr;
 
-    shared_ptr(BlockAndData<T>* cb) : cb_(cb) { ptr_ = cb->getObject(); };
+    shared_ptr(BlockAndData<T>* cb) : cb_(cb) { ptr_ = new T{cb->getObject()}; };
```
ziobron, line 40: no, you have 2 objects right now. In a control block and a newly created on the heap.

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -110,8 +101,9 @@ shared_ptr<T>& shared_ptr<T>::operator=(shared_ptr&& ptr) {
 
 template <typename T>
 void shared_ptr<T>::reset(T* ptr) {
-    *ptr_ = *ptr;
-    delete ptr;
+    deletePointers();
+    ptr_ = ptr;
+    cb_ = new ControlBlock<T>;
```
ziobron, line 106: and decrement shared refs if ptr_ is nullptr. Add a proper test for it. use_count should return 0 for nulltprs.

homework/shared_ptr/shared_ptr.hpp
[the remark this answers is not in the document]
ziobron: ok

homework/shared_ptr/shared_ptr.hpp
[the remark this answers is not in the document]
ziobron: No. it will be called on the pointer that was set during the construction of the control block, and it was exactly ptr.ptr_, which you have just copied to ptr_. So it should be ok.

homework/shared_ptr/weak_ptr-ut.cpp
```diff
@@ -34,35 +34,41 @@ TEST(WeakPointerTests, WeakPointerLock) {
...
-    constexpr size_t numberOfUseCount = 1;
+    size_t numberOfUseCount = 0;
     struct Something {};
     cs::shared_ptr<Something> shared_something(new Something{});
     cs::weak_ptr<Something> weak_something{shared_something};
     cs::weak_ptr<Something> weak_ptr;
+    ASSERT_EQ(weak_ptr.use_count(), numberOfUseCount);
     weak_ptr = std::move(weak_something);
-
+    ++numberOfUseCount;
```
ziobron, line 44: I really don't like the incrementation here and in a few other places. It tells me that the first the other use_count has something in common. In fact, they are independent, because they use another share_ptr. I opt for having two separated numberOfUseCount variables with 0 and 1 values here.

homework/shared_ptr/weak_ptr-ut.cpp
```diff
@@ -34,35 +34,41 @@ TEST(WeakPointerTests, WeakPointerLock) {
 }
 
 TEST(WeakPointerTests, moveAssignmentOperatorShouldMovePointer) {
-    constexpr size_t numberOfUseCount = 1;
+    size_t numberOfUseCount = 0;
     struct Something {};
     cs::shared_ptr<Something> shared_something(new Something{});
     cs::weak_ptr<Something> weak_something{shared_something};
     cs::weak_ptr<Something> weak_ptr;
```
ziobron, line 41: please initialize this weak_ptr with another shared_ptr before move. With such empty weak_ptr, some resources may not leak.

homework/shared_ptr/weak_ptr-ut.cpp
```diff
@@ -34,35 +34,41 @@ TEST(WeakPointerTests, WeakPointerLock) {
...
 }
 
-TEST(WeakPointerTests, copyAssignmentOperatorShouldMovePointer) {
-    constexpr size_t numberOfUseCount = 1;
+TEST(WeakPointerTests, copyAssignmentOperatorShouldCopyPointer) {
+    size_t numberOfUseCount = 0;
     struct Something {};
     cs::shared_ptr<Something> shared_something(new Something{});
     cs::weak_ptr<Something> weak_something{shared_something};
     cs::weak_ptr<Something> weak_ptr;
```
ziobron, line 53: please initialize this weak_ptr with another shared_ptr before copy. With such empty weak_ptr, some resources may not leak.

homework/shared_ptr/weak_ptr-ut.cpp
```diff
@@ -34,35 +34,41 @@ TEST(WeakPointerTests, WeakPointerLock) {
...
-  weakSomething.reset();
-  weakSomething = next_ptr;
-  auto locked = weakSomething.lock();
-  ASSERT_EQ(expectedValue, *locked);
+    constexpr size_t numberOfUseCount = 0;
+    int expectedValue{100};
+    auto s_ptr = cs::shared_ptr<int>(new int{20});
+    auto next_ptr = cs::shared_ptr<int>(new int{expectedValue});
+    cs::weak_ptr<int> weakSomething{s_ptr};
+    weakSomething.reset();
```
ziobron, line 68: you can also check that this reset does not change s_ptr and everything is okay with it. Also, check use_count before reset.

ziobron (changes requested):

homework/shared_ptr/control_block.hpp
Kuba-Wi: I don't want to change so it's const&. It doesn't return pointer because object_ is allocated on stack and in place where I call this method I allocate memory via new. Ok I could return pointer and then dereference it to call new, but it would be pointless.

homework/shared_ptr/shared_ptr.hpp
Kuba-Wi: Previous deleter was faulty.

homework/shared_ptr/shared_ptr.hpp
Kuba-Wi: But one is created on the stack.

homework/shared_ptr/shared_ptr-ut.cpp
```diff
@@ -36,47 +39,53 @@ TEST(sharedPointerTest, MovingContructorShouldMovePointer) {
...
-    ++numberOfUseCount;
-    ASSERT_EQ(copied_ptr.use_count(), numberOfUseCount);
+TEST(sharedPointerTest, resetMethodWithoutArgumentsShouldChangePointedObject) {
+    size_t numberOfUseCount = 2;
+    cs::shared_ptr<int> ptr(new int{0});
+    cs::shared_ptr<int> ptr_doing_nothing(ptr);
+    ASSERT_EQ(ptr.use_count(), numberOfUseCount);
+    ptr.reset();
+    numberOfUseCount = 1;
     ASSERT_EQ(ptr.use_count(), numberOfUseCount);
```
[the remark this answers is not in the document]
Kuba-Wi: I checked for std::shared_ptr and use_count before reset was 2.

homework/shared_ptr/shared_ptr.hpp
Kuba-Wi: When I copy shared_ptr that points on nullptr then delete on cb_ is called twice. So more convenient is to leave it as it is and add condition in use_count().

homework/shared_ptr/shared_ptr-ut.cpp
```diff
@@ -36,47 +39,53 @@ TEST(sharedPointerTest, MovingContructorShouldMovePointer) {
...
 }
 
 TEST(makeSharedTest, makeSharedCreatesTheSameValueAsSharedPtr) {
-    cs::shared_ptr<int> someInt(new int{42});
-    auto makeSomeInt = cs::make_shared<int>(42);
+    constexpr int value = 42;
+    cs::shared_ptr<int> someInt(new int{value});
+    auto makeSomeInt = cs::make_shared<int>(value);
     ASSERT_EQ(*makeSomeInt, *someInt);
 }
```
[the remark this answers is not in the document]
Kuba-Wi: added tests for copying and moving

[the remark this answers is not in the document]
kuchichan: Also, additional test for `tuple<int, int, int>` 3 parameter and empty string (0 parameter) added.

homework/shared_ptr/shared_ptr-ut.cpp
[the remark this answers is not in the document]
ziobron: that's true. But my comment was about use_count after reset, not before.
ptr.use_count = 0
ptr_doing_nothing.use_count = 1

ziobron (approved):
Good job 👍

ziobron:
{
  "requirement_ids": [3794, 3795, 3796],
   "users": ["kuchichan", "Kuba-Wi"]
}

ziobron:
{
"requirement_ids": [1636],
"users": ["kuchichan", "Kuba-Wi"]
}

PR #60, opened 2021-09-26
Passing char 'd', throwing exception,  memory leak

Regggis:
After passing pointer to char 'd', there is problem with memory leak. Don't know exactly why, because we are deleting rsc twice?


ziobron:
✅ CI Passed - Task: resource
🏅 Granted 5 XP for @Regggis

ziobron:
⏱ 3 XP extra for @Regggis for on-time delivery

ziobron:
Because we do not delete the memory. ‘delete src’ is not called because of the exception that is thrown when we pass ‘d’

PR #80, opened 2021-10-20
unique_ptr homework

ploWoj:


nauka-programowania-MA (approved):
Nice one!

homework/shared_ptr/controlBlock.hpp
```diff
@@ -0,0 +1,83 @@
...
+} 
+
+template <typename T>
+void ControlBlock<T>::decreaseWeakRef()
+{
+    if (weakRef_ > 0)
+    {
+        weakRef_--;
+    }
+};
```
nauka-programowania-MA, line 63: control block should delete resource not shared_ptr

homework/shared_ptr/controlBlock.hpp
```diff
@@ -0,0 +1,83 @@
...
+template <typename T>
+void ControlBlock<T>::decreaseSharedRef()
+{
+    if (sharedRef_ > 0)
+    {
+        sharedRef_--;
+    }
+};
```
nauka-programowania-MA, line 71: Control block should delete himself when both ref drop to 0. shared_ptr should not manage it!

homework/shared_ptr/controlBlock.hpp
nauka-programowania-MA: also Control block should delete himself when both ref drop to 0. shared_ptr should not manage it!

nauka-programowania-MA: I also noticed that you want to create it thread safe. So here you should use mutex. because both counter could be modified by other threads. Making it atomic is not enough. Better work on normal int and use mutex. Or hold both references into one variable (and youse bit shifting, first 32 but is strong ref rest of variable is weak_ref). You can also make a struct with size 8 bytes and 2 fields each 4 bytes. And then you can atomic modify it

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+#include "controlBlock.hpp"
+
+class weak_ptr;
+
+template <typename T>
+class shared_ptr
+{
+public:
+    shared_ptr(T *ptr){};
```
nauka-programowania-MA, line 13: = default

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,112 @@
...
+
+    // Copy constructor
+    shared_ptr(const shared_ptr &ptr) noexcept;
+    shared_ptr(shared_ptr &&ptr) noexcept;
+    ~shared_ptr() noexcept;
+
+    shared_ptr<T> &operator=(const shared_ptr &ptr) noexcept;
+    shared_ptr<T> &operator=(shared_ptr &&ptr) noexcept;
+
+    T &operator*() const noexcept { return *ptr_; };
```
nauka-programowania-MA, line 23: dereference nullptr can throw an error

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,112 @@
...
+    T *operator->() const noexcept { return ptr_; };
+    const T *get() const noexcept;
```
nauka-programowania-MA, line 25: You need bot version for const method (return const object) and for non const return non const

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,112 @@
...
+    T *reset(T* ptr) noexcept;
+    long use_count() const noexcept;
+    explicit operator bool() const noexcept;
+
+private:
+    T *rawPtr_{nullptr};
+    ControlBlock<T> *cb_{nullptr};
+    void releaseMemory();
```
nauka-programowania-MA, line 33: this should manage control block

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,112 @@
...
+template <typename T>
+const T *shared_ptr<T>::get() const noexcept
+{
+    return rawPtr_;
+}
+
+template <typename T>
+T *shared_ptr<T>::reset(T* ptr) noexcept
+{
+    releaseMemory();
```
nauka-programowania-MA, line 94: Just decrease strong ref. Control block will check if this is the last ptr and delete it. You can reset one shared_p[tr but others are still alive, so you don't want to release resources.

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,112 @@
...
+    if (cb_)
+    {
+        return cb_->getSharedRef();
+    }
+    return 0;
+}
+
+template <typename T>
+explicit shared_ptr<T>::operator bool() const noexcept {
+    return ptr_ != nullptr;
```
nauka-programowania-MA, line 111: just return ptr_

homework/shared_ptr/shared_ptr.hpp
```diff
@@ -0,0 +1,112 @@
...
+class weak_ptr;
+
+template <typename T>
+class shared_ptr
+{
+public:
+    shared_ptr(T *ptr){};
+
+    // Copy constructor
+    shared_ptr(const shared_ptr &ptr) noexcept;
```
nauka-programowania-MA, line 16: Shared_ptr should have also a constructor for wak_ptr. You can create shared_ptr form weak_ptr :)

homework/shared_ptr/weak_ptr.hpp
```diff
@@ -0,0 +1,108 @@
...
+template <typename T>
+class Weak_Ptr
+{
+private:
+    T *rawPtr_;
+    ControlBlock<T> *cb_;
+    void releaseMemory();
+
+public:
+    Weak_Ptr() = default;
```
nauka-programowania-MA, line 15: Weak ptr should be constructed from shared_ptr!

homework/shared_ptr/weak_ptr.hpp
```diff
@@ -0,0 +1,108 @@
...
+Weak_Ptr<T>::Weak_Ptr(Weak_Ptr &&ptr) noexcept : rawPtr_(ptr.rawPtr_), cb_(ptr.cb_)
+{
+    ptr.rawPtr_ = nullptr;
+    ptr.cb_ = nullptr;
+}
+
+template <typename T>
+Weak_Ptr<T>::~Weak_Ptr()
+{
+    releaseMemory();
```
nauka-programowania-MA, line 45: no just decrase weak counter. Control block will manage resource

homework/shared_ptr/weak_ptr.hpp
```diff
@@ -0,0 +1,108 @@
...
+template <typename T>
+void Weak_Ptr<T>::releaseMemory() {
+    if (cb_) {
+        cb_->decreaseWeakRef();
+        if (cb_->getSharedRef() == 0 && cb_->getWeakRef() == 0)
+        {
+            delete cb_;
+            cb_ = nullptr;
+        }
+    }
```
nauka-programowania-MA, line 57: remove

homework/shared_ptr/weak_ptr.hpp
```diff
@@ -0,0 +1,108 @@
...
+}
+
+template <typename T>
+Weak_Ptr<T> &Weak_Ptr<T>::operator=(const Weak_Ptr &ptr) noexcept
+{
+    if (ptr != this)
+    {
+        releaseMemory();
```
nauka-programowania-MA, line 65: just decrease weak_counter here

homework/shared_ptr/weak_ptr.hpp
```diff
@@ -0,0 +1,108 @@
...
+long Weak_Ptr<T>::use_count() const noexcept {
+    if (cb_) {
+        return cb_->getSharedRef();
+    }
+    return 0;
+}
+
+template <typename T>
+shared_ptr<T> Weak_Ptr<T>::lock() const noexcept {
+    return expired() ? shared_ptr<T>() : shared_ptr<T>(*this);
```
nauka-programowania-MA, line 100: return nullptr when expired

PR #95, opened 2022-01-07
Simple Unique_ptr

NikczemnyOkularnik:
@Kasia9311  @Crazyayane07  @NikczemnyOkularnik

### DIFF
--- a/homework/unique_ptr/test/test.cpp
+++ b/homework/unique_ptr/test/test.cpp
@@ -1,0 +1,113 @@
+#define CATCH_CONFIG_MAIN // This tells Catch to provide a main() - only do this in one cpp file
+#include "catch_amalgamated.hpp"
+#include "../uniquePTR.hpp"
+#include <type_traits>
+
+SCENARIO("Testing unique pointer - copy")
+{
+    GIVEN("Two scenarios")
+    {
+        WHEN("Checking if it is possible to copy by copy constructor")
+        {
+            auto copy = std::is_copy_constructible<unique_ptr<int>>::value;
+            THEN("Result should be false")
+            {
+                REQUIRE_FALSE(copy);
+            }
+        }
+        WHEN("Checking if it is possible to copy by copy assignment operator")
+        {
+            auto copy2 = std::is_copy_assignable<unique_ptr<int>>::value;
+            THEN("Result should be false")
+            {
+                REQUIRE_FALSE(copy2);
+            }
+        }
+    }
+}
+
+SCENARIO("Testing unique pointer - move")
+{
+    GIVEN("Two scenarios")
+    {
+        WHEN("Checking if it is possible to move by move constructor")
+        {
+            auto move = std::is_move_constructible<unique_ptr<int>>::value;
+            THEN("Result should be true")
+            {
+                REQUIRE(move);
+            }
+        }
+        WHEN("Checking if it is possible to move by move assignment operator")
+        {
+            auto move2 = std::is_move_assignable<unique_ptr<int>>::value;
+            THEN("Result should be true")
+            {
+                REQUIRE(move2);
+            }
+        }
+    }
+}
+
+SCENARIO("Testing unique pointer - operators")
+{
+    GIVEN("Two scenarios")
+    {
+        WHEN("Checking overloading operator*()")
+        {
+            auto unique1 = unique_ptr<int>(new int (5));
+            auto operator1 = unique1.operator*();
+            THEN("Value of operator1 should be equal to 5")
+            {
+                CHECK(operator1 == 5);
+            }
+        }
+        WHEN("Checking overloading operator->()")
+        {
+            auto unique2 = unique_ptr<int>(new int (17));
+            auto operator2 = unique2.operator->();
+            THEN("Result should be true")
+            {
+                CHECK(*operator2 == *unique2);
+            }
+        }
+    }
+}
+
+SCENARIO("Testing unique pointer - functions")
+{
+    GIVEN("3 scenarios")
+    {
+        WHEN("Checking function get()")
+        {
+            auto unique1 = unique_ptr<int>(new int (5));
+            auto operator1 = unique1.get();
+            THEN("Value of operator1 should be equal to 5")
+            {
+                CHECK(*operator1 == 5);
+            }
+        }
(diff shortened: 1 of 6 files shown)


homework/unique_ptr/uniquePTR.hpp
```diff
@@ -0,0 +1,55 @@
...
+    ptr_ = rhs.ptr_;
+    rhs.ptr_ = nullptr;
+  }
+
+  unique_ptr& operator=(const unique_ptr &) = delete;
+  unique_ptr& operator=(unique_ptr && rhs){
+    if (this != &rhs) {
+      delete ptr_;
+      ptr_ = rhs.ptr_;
+      rhs.ptr_ = nullptr;
```
WojtekMs, line 23: These two lines can be collapsed to `ptr_ = std::exchange(rhs.ptr_, nullptr);`

homework/unique_ptr/uniquePTR.hpp
```diff
@@ -0,0 +1,55 @@
...
+  unique_ptr() = default;
+  explicit unique_ptr(T* ptr) : ptr_(ptr){}
+  unique_ptr(const unique_ptr &) = delete;
+  unique_ptr(unique_ptr&& rhs) {
+    ptr_ = rhs.ptr_;
+    rhs.ptr_ = nullptr;
+  }
+
+  unique_ptr& operator=(const unique_ptr &) = delete;
+  unique_ptr& operator=(unique_ptr && rhs){
```
WojtekMs, line 19: Same comment as to move constructor

homework/unique_ptr/uniquePTR.hpp
```diff
@@ -0,0 +1,55 @@
...
+template <typename T>
+class unique_ptr 
+{
+  T* ptr_{nullptr};
+
+public:
+  unique_ptr() = default;
+  explicit unique_ptr(T* ptr) : ptr_(ptr){}
+  unique_ptr(const unique_ptr &) = delete;
+  unique_ptr(unique_ptr&& rhs) {
```
WojtekMs, line 13: Move operations should be atomic (non-dividable), so there shouldn't be any exceptions thrown. It is worth using `noexcept` specifier.

homework/unique_ptr/uniquePTR.hpp
```diff
@@ -0,0 +1,55 @@
...
+    ptr_ = rhs.ptr_;
+    rhs.ptr_ = nullptr;
```
WojtekMs, line 15: These two lines can be collapsed to `ptr_ = std::exchange(rhs.ptr_, nullptr);`. It can be done on the initializer list :)

homework/unique_ptr/uniquePTR.hpp
```diff
@@ -0,0 +1,55 @@
+#pragma once
+#include <iostream>
```
WojtekMs, line 2: Avoid unnecesary headers, they increase compilation time :)

homework/unique_ptr/uniquePTR.hpp
```diff
@@ -0,0 +1,55 @@
...
+    if (this != &rhs) {
+      delete ptr_;
+      ptr_ = rhs.ptr_;
+      rhs.ptr_ = nullptr;
+    }
+
+    return *this;
+  }
+
+  T& operator*(){
```
WojtekMs, line 29: This method can be const

homework/unique_ptr/uniquePTR.hpp
```diff
@@ -0,0 +1,55 @@
...
+      return *ptr_;
+  }
+
+  T* operator->(){
```
WojtekMs, line 33: This method can be const

homework/unique_ptr/uniquePTR.hpp
```diff
@@ -0,0 +1,55 @@
...
+      return ptr_;
+  }
+
+  T* get(){
```
WojtekMs, line 37: This method can be const

homework/unique_ptr/test/test.cpp
```diff
@@ -0,0 +1,113 @@
...
+            }
+        }
+    }
+}
+
+SCENARIO("Testing unique pointer - move")
+{
+    GIVEN("Two scenarios")
+    {
+        WHEN("Checking if it is possible to move by move constructor")
```
WojtekMs, line 33: In my opinion this is not really enough to check if it is possible. You should try to check if it works correctly (try to use the function and verify with valgrind that there are no leaks ;))

WojtekMs (commented):
Good work overall :)

ziobron:
Thanks @WojtekMs for Code Review.
🏅 1 XP granted

homework/unique_ptr/uniquePTR.hpp
```diff
@@ -0,0 +1,55 @@
...
+      return ptr_;
+  }
+
+  T* release(){
+      T* newPtr = ptr_;
+      ptr_ = nullptr;
+      return newPtr;
```
WojtekMs, line 44: I think that these 3 lines can also be collapsed to single line: `return std::exchange(ptr_, nullptr);`

ziobron:
Thanks @WojtekMs for Code Review.
🏅 1 XP granted

homework/unique_ptr/test/test.cpp
```diff
@@ -0,0 +1,114 @@
...
+        }
+        WHEN("Checking if it is possible to copy by copy assignment operator")
+        {
+            auto copy2 = std::is_copy_assignable<unique_ptr<int>>::value;
+            THEN("Result should be false")
+            {
+                REQUIRE_FALSE(copy2);
+            }
+        }
+    }
```
nauka-programowania-MA, line 26: this could be done by static_assert inside unique_ptr class. During compilation, all conditions will be checked. Making test for type traits is redundend

homework/unique_ptr/test/test.cpp
nauka-programowania-MA: Agree. Don't test type_traits. Test real usage. Move this pointer use Move C'tor and move operator= and check if there is no memory leak :)

nauka-programowania-MA (approved):
Nice :) You provide noexcept for move semantic, and use std::change :)

PR #96, opened 2022-01-07
valgrind-output.txt added

pwiatr3:


ziobron (changes requested):
The script has expected this file to be in homework/resourceD/valgrind-output.txt. Please move it and then everything should be ok.

pwiatr3:
fail

PR #100, opened 2022-01-10
FIX: check.sh

### LINKED ISSUE #99: A testcase could be more general
On brunch `resource` in `memory-management/.github/scripts/check.sh` one testcase assumes that user didn't modify the file before pushing to the server:
`pattern ${FILE} "Looking for the allocation place" "main \(resourceD.cpp:30\)"`

michalsoloducha:
Fix #99


ziobron (approved):

ziobron:
Your PR was merged!
🏅 2 XP granted. Thanks @michalsoloducha for making course materials better!

PR #115, opened 2022-03-26
Running student's tests

anraf1001:
GitHub Actions jobs that runs shared_ptr, weak_ptr and make_shared unit tests written by student using GoogleTest or Catch2


tests/shared_ptr_student/cmake_config.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
ziobron, line 1: it's the same file as the rest of cmake_config.sh.
Please move it to `.github/scripts` and take `$file` as an argument, because it's the only thing that differs.

tests/make_shared_student/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
+cmake_minimum_required(VERSION 3.14)
```
ziobron, line 1: Please merge all CMakeLists.txt into a single file and put it in `.github/scripts`
Just give different names to each executable.
Then in main.yml just select a target using executable_name from `add_executable` ane run it like: `make shared_ptr_tests`

ziobron (changes requested):

ziobron (approved):

ziobron:
Your PR was merged!

PR #118, opened 2022-03-29
Checking if shared_ptr interface is fully implemented

anraf1001:


tests/shared_ptr_interface_check.sh
```diff
@@ -0,0 +1,55 @@
...
+    fi
+
+    return 0
+}
+
+file=../homework/shared_ptr/shared_ptr.hpp
+exit_status=0
+echo "Checking if shared_ptr interface is fully implemented..."
+
+check_regex $file 'shared_ptr\s*\(const shared_ptr& \w+\)' 'Copy constructor'
```
ziobron, line 19: What if some put `shared_ptr(shared_ptr const &)`? or `shared_ptr(const shared_ptr & sth)`?
It may be easier to first apply a clang-format rules before running those checks. Otherwise we also need to handle all types of formatting here. What if someone is gonna put args in the next line?
```cpp
shared_ptr(
    shared_ptr const & item
);
```

tests/shared_ptr_interface_check.sh
```diff
@@ -0,0 +1,55 @@
...
+exit_status=$(($exit_status | $?))
+
+check_regex $file 'shared_ptr\s*\(shared_ptr&& \w+\)' 'Move constructor'
```
ziobron, line 22: What about `shared_ptr(shared_ptr &&item)`?

ziobron (changes requested):

.github/workflows/main.yml
```diff
@@ -113,3 +113,15 @@ jobs:
...
+  check_shared_ptr_interface:
+    needs: shared_ptr_formatting_check
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Check shared_ptr has whole interface implemented
+      working-directory: tests/
+      run: bash ./shared_ptr_interface_check.sh
```
ziobron, line 127: maybe just 
```suggestion
      run: ./shared_ptr_interface_check.sh
```

tests/shared_ptr_interface_check.sh
```diff
@@ -0,0 +1,55 @@
...
+    fi
+
+    return 0
+}
+
+file=../homework/shared_ptr/shared_ptr.hpp
+exit_status=0
+echo "Checking if shared_ptr interface is fully implemented..."
+
+check_regex $file 'shared_ptr\s*\(\n?\s*?((const shared_ptr& \w+)|(shared_ptr const& \w+))\n?\s*?\)' 'Copy constructor'
```
ziobron, line 19: Please add some tests showing which outputs are accepted and which are not. Take a look here: https://github.com/coders-school/git-checker/tree/main/.github/scripts, You can reuse some scripts or functions from there.
It would be good to have them split for each requirement (function like copy constructor, move constructor etc.) in a separate file.

tests/shared_ptr_interface_check.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
ziobron, line 1: Please move it into .github/scripts

ziobron (changes requested):

.github/scripts/shared_ptr_interface_check.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+source functions.sh
+
+FILE=../../homework/shared_ptr/shared_ptr.hpp
```
ziobron, line 5: I'm never 100% sure about using " in scripts, but I guess that it is safe to put paths in quotes.
```suggestion
FILE="../../homework/shared_ptr/shared_ptr.hpp"
```

.github/scripts/functions.sh
```diff
@@ -0,0 +1,121 @@
+#!/bin/bash
+
+failed=0
+total=0
+
+# 1: file, 2: regex, 3: method_to_find
+function check_regex () {
+    (( total++ ))
+
+    grep -zoP "$2" $1 > /dev/null
```
ziobron, line 10: why `-zoP`? What does it do? I don't even have it in `man grep`
Put `$1` in quotes as well.

.github/scripts/functions.sh
```diff
@@ -0,0 +1,121 @@
...
+    fi
+
+    echo "✅ $3 is implemented"
+    return 0
+}
+
+# 1: file
+function check_shared_ptr_copy_constructor() {
+    NAME='Copy constructor'
+    PATTERN='shared_ptr\s*\(\n?\s*?((const shared_ptr)|(shared_ptr const))&[^&]*?\)'
```
ziobron, line 25: Do not put spaces in regex for C++ code :) `\s+` is more generic and can behave properly if someone put an additional space.
`\s` already contain `\n`, there is no need to have it in a regex. 
`*?` doesn't make sense. `*` means any number, including 0.
```suggestion
    PATTERN='shared_ptr\s*\(\s*((const\s+shared_ptr)|(shared_ptr\s+const))\s*&.*\)'
```
Please check if my changes still pass the tests.
You can test your regex using https://regex101.com

.github/scripts/tests.sh
```diff
@@ -0,0 +1,273 @@
...
+
+source functions.sh
+
+function test_shared_ptr_copy_constructor() {
+    declare -a right_copy_ctors=("shared_ptr(const shared_ptr& other)"
+                                 "shared_ptr(shared_ptr const& ptr)"
+                                 "shared_ptr(const shared_ptr&)"
+                                 "shared_ptr(shared_ptr const&)"
+                                 "shared_ptr(
+                                     const shared_ptr& other)")
```
ziobron, line 11: what about 
`shared_ptr(shared_ptr const &sth)`
`shared_ptr(shared_ptr const & sth)`
`shared_ptr(const shared_ptr &sth)`
`shared_ptr(const shared_ptr & sth)`

I know that clang-format can unify that, but probably it's a really small change in regex. Thanks to that we can be more independent from clang-format.
Please also add such cases. I hope that my modification of regex can make it work.

.github/scripts/tests.sh
```diff
@@ -0,0 +1,273 @@
...
+
+    for el in "${right_copy_ctors[@]}"; do
+        echo $el | check_shared_ptr_copy_constructor /dev/stdin > /dev/null
```
ziobron, line 14: why not 
```suggestion
        check_shared_ptr_copy_constructor "$el" > /dev/null
```

.github/scripts/tests.sh
```diff
@@ -0,0 +1,273 @@
...
+        echo $el | check_shared_ptr_move_constructor /dev/stdin > /dev/null
+        if [[ $? -eq 0 ]]; then
+            echo "❌ Wrong move ctors check not passed: $el"
+            return 1
+        fi
+    done
+
+    echo "✅ Move ctor check passed"
+    return 0
+}
```
ziobron, line 67: There is a lot of code duplication here. Extract common parts to function.
I can see 3 variables here - `right_patterns`, `wrong_patterns`, `check_function`

ziobron (changes requested):

.github/scripts/tests.sh
anraf1001: `egrep` needs a file path as parameter, not a string. I don't have an idea how to make it clearer without making interface checking function more complicated (e.g. passing content of file to `check_regex` function and then piping it into `egrep`)

.github/scripts/tests.sh
```diff
@@ -2,263 +2,269 @@
...
-        echo $el | check_shared_ptr_move_constructor /dev/stdin > /dev/null
-        if [[ $? != 0 ]]; then
-            echo "❌ Right move ctors check not passed: $el"
-            return 1
-        fi
-    done
+    check_right_cases $1 right_cases[@]
+    if [[ $? != 0 ]]; then
+        ret=1
+    fi
```
ziobron, line 43: why not `ret = $?` without if?

.github/scripts/tests.sh
```diff
@@ -2,263 +2,269 @@
...
+    run_test check_shared_ptr_copy_constructor right_copy_ctors[@] wrong_copy_ctors[@]
+    if [[ $? != 0 ]]; then
+        echo "❌ Copy ctor check not passed"
+        echo
+        return 1
+    fi
+
+    echo "✅ Copy ctor check passed"
+    echo
     return 0
```
ziobron, line 81: You can move this common section to `run_tests`. Just pass an additional string to display which check passed or not.

.github/scripts/tests.sh
```diff
@@ -2,263 +2,269 @@
...
-            return 1
-        fi
-    done
+    local right_dereference_operator=("T& operator*()"
+                                      "Type& operator*()")
 
-    declare -a wrong_dereference_operator=("T&& operator*()"
-                                           "T operator*()")
+    local wrong_dereference_operator=("T&& operator*()"
+                                      "T operator*()")
```
ziobron, line 175: I would like to see a test with different type name and additional spaces like 
`MyType & operator *()`
`N &operator* ()`
It should go to all template methods.

.github/scripts/tests.sh
```diff
@@ -2,263 +2,269 @@
...
-        if [[ $? != 0 ]]; then
-            echo "❌ Right use_count method check not passed: $el"
-            return 1
-        fi
-    done
+    local right_use_count=("size_t use_count()"
+                           "long use_count()")
 
-    declare -a wrong_use_count=("size_t use_count(shared_ptr ptr)")
+    local wrong_use_count=("size_t use_count(shared_ptr ptr)")
```
ziobron, line 238: what about
`int use_count()`
?

ziobron (changes requested):

.github/scripts/tests.sh
ziobron: ok, it can stay like that.

.github/scripts/tests.sh
anraf1001: Then `wrong_cases`'s exit code will override `right_cases`'s one.
I can see 2 options to avoid that if statement:
1. Just adding `$?` to `ret` without caring if `ret` equals 1 or 2
2. Using bitwise or (`ret=$(( $ret | 1 ))`) if we need only 2 values of exit (0 and 1)

.github/scripts/tests.sh
```diff
@@ -0,0 +1,280 @@
+#!/bin/bash
+
+source functions.sh
+
+function check_right_cases() {
```
ziobron, line 5: can you add comments with the description of function parameters?

.github/scripts/tests.sh
```diff
@@ -0,0 +1,280 @@
...
+    return $ret
+}
+
+function check_wrong_cases() {
+    declare -a wrong_cases=("${!2}")
+    local ret=0
+
+    for el in "${wrong_cases[@]}"; do
+        echo $el | $1 /dev/stdin > /dev/null
+        if [[ $? -eq 0 ]]; then
```
ziobron, line 26: in line 11 there is `!=`. Here we have `-eq`. Please use only one uniform way.

.github/scripts/tests.sh
ziobron: Maybe just sum the return values. If we have 0 then everything is fine. Anything else means that there was an error.

.github/scripts/tests.sh
```diff
@@ -0,0 +1,280 @@
...
+                            "shared_ptr(const shared_ptr &sth)"
+                            "shared_ptr(const shared_ptr & sth)")
+
+    local wrong_copy_ctors=("shared_ptr(const shared_ptr&& other)"
+                            "shared_ptr(shared_ptr&&)"
+                            "shared(const shared_ptr& other)"
+                            "shared_ptr(shared_ptr& ptr)")
+
+    run_test "copy ctor" check_shared_ptr_copy_constructor right_copy_ctors[@] wrong_copy_ctors[@]
+    return $?
```
ziobron, line 80: can we merge those 2 lines?

```suggestion
    return run_test "copy ctor" check_shared_ptr_copy_constructor right_copy_ctors[@] wrong_copy_ctors[@]
```

or the behavior is different then?

.github/scripts/tests.sh
```diff
@@ -0,0 +1,280 @@
...
+    run_test "dereference operator" check_shared_ptr_dereference_operator right_dereference_operator[@] wrong_dereference_operator[@]
+    return $?
+}
+
+function test_shared_ptr_arrow_operator() {
+    local right_arrow_operator=("T* operator->()"
+                                "Type* operator->()"
+                                "MyType* operator -> ()"
+                                "MyType * operator->()"
+                                "N *operator-> ()")
```
ziobron, line 157: Are all of these cases problematic, or can we only comment out specific lines here, if some of them are ok?

ziobron (changes requested):

.github/scripts/tests.sh
anraf1001: As far as I know, `return` in bash can only return exit codes, which can only be accessed using `$?`. Suggested version won't work at all.

.github/scripts/tests.sh
```diff
@@ -260,21 +259,51 @@ function test_use_count_usage() {
...
+
 test_get_usage
+ret=$(( $ret + $? ))
+
 test_reset_usage
+ret=$(( $ret + $? ))
+
 test_use_count_usage
+ret=$(( $ret + $? ))
+
```
ziobron, line 308: This code with repetitions of summing up return value looks ugly :)
Can you just use a global variable like `error_count` that is incremented in `run_test()` or `check_wrong_cases ()` or `check_right_cases()` functions and simply return it here?
In scripts, it is not a big deal to use global variables as they usually have the scope of one line and are not compiled from multiple source files.

.github/workflows/main.yml
```diff
@@ -153,3 +153,35 @@ jobs:
...
+    - uses: actions/upload-artifact@v3
+      if: failure()
+      with:
+        name: Coverage-report
+        path: .github/scripts/build/report/
+    - uses: JoseThen/comment-pr@v1.1.1
+      if: ${{ failure() && github.event_name == 'pull_request' }}
+      with:
+        comment: "Coverage rate is less than 90%. You can download detailed coverage report in action's summary"
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
ziobron, line 187: This should be a separate PR. It has nothing in common with checking if the shared_ptr interface is fully implemented. Please raise a new one with this change.

ziobron (changes requested):
Recently I have refreshed my knowlege on good pracices regarding commit messages. Please read and stick to this doc -https://reflectoring.io/meaningful-commit-messages/
I know that my previous commits does not stick to it, but now I also use these rules :)

ziobron (approved):

ziobron:
Your PR was merged!

PR #120, opened 2022-04-07
unique_ptr tests and clang-format check

anraf1001:
* `unique_ptr` tests
* running tests added by student
* clang-format check


ziobron (changes requested):
Please extract changes in shared_ptr files to a separate PR (it can go to the existing PR with shared_ptr). They should be independent or at least be in separate commits.
Also, a good practice is to have one logical change in one commit. Eg. here I would see 3 commits - 1. unque_ptr_tests, 2. running students' tests, 3. clang-format.

I have created a separate branch named `unique_ptr` for unique_ptr CI. It should contain only `homework/unique_ptr` dir and `.github` dir. Please remove anything else (as a separate commit) and then apply your work. It would be much easier to run only appropriate jobs on branches.

When we finish work for shared_ptr I would like to move it to `shared_ptr` branch as well.

.github/scripts/tests/unique_ptr_tests.cpp
```diff
@@ -0,0 +1,62 @@
...
+    ASSERT_EQ(*ptr, initialValue);
+}
+
+TEST_F(UniquePtrTest, shouldMoveUniquePtr) {
+    auto ptr2{std::move(ptr)};
+
+    EXPECT_THAT(ptr.get(), IsNull());
+    ASSERT_EQ(*ptr2, initialValue);
+
+    my::unique_ptr<int> ptr3{};
```
ziobron, line 31: you can add checking if ptr3 is a nullptr after creation

.github/scripts/tests/unique_ptr_tests.cpp
```diff
@@ -0,0 +1,62 @@
...
+    ptr3 = std::move(ptr2);
+
+    EXPECT_THAT(ptr2.get(), IsNull());
+    ASSERT_EQ(*ptr3, initialValue);
+}
+
+TEST_F(UniquePtrTest, shouldUseResetMethod) {
+    constexpr int newValue = 6;
+    ptr.reset(new int{newValue});
```
ziobron, line 40: Add a test for `reset()` with no arguments

PR #126, opened 2022-04-15
Checking if unique_ptr interface is fully implemented

anraf1001:


.github/scripts/tests.sh
```diff
@@ -0,0 +1,296 @@
...
+ret=$(( $ret + $? ))
+
+test_get_usage
+ret=$(( $ret + $? ))
+
+test_reset_usage
+ret=$(( $ret + $? ))
+
+test_release_usage
+ret=$(( $ret + $? ))
```
ziobron, line 296: The same issue here as for shared_ptrs - https://github.com/coders-school/memory-management/pull/118#discussion_r859993718
This code with repetitions of summing up return value looks ugly :)
Can you just use a global variable like error_count that is incremented in run_test() or check_wrong_cases () or check_right_cases() functions and simply return it here?
In scripts, it is not a big deal to use global variables as they usually have the scope of one line and are not compiled from multiple source files.

.github/scripts/tests.sh
```diff
@@ -43,19 +45,18 @@ function run_test() {
...
     check_wrong_cases $2 wrong_cases[@]
-    ret=$(( $ret + $? ))
+    error_count=$(( $error_count + $? ))
 
-    if [[ $ret != 0 ]]; then
+    if [[ $error_count != 0 ]]; then
         echo "❌ $1 check not passed"
     else
         echo "✅ $1 check passed"
     fi
```
ziobron, line 57: It's not gonna work as intended right now. After the first error, it will display "check not passed" for everything else.
I think it's not needed here. Remove this whole block and we are done.
```suggestion
```

ziobron (changes requested):

ziobron (approved):

ziobron:
Your PR was merged!

PR #140, opened 2022-04-24
Task completed

mBialczak:
The error occurence depends on the input given:
when an argument starting with 'd' is passed,
an exception is thrown, causing stack unwinding
and delete is not called.


ziobron:
✅ CI Passed - Task: resource
🏅 Granted 5 XP for @mBialczak

PR #153, opened 2022-05-04
Info about low coverage rate - shared_ptr

anraf1001:


.github/scripts/coverage.sh
```diff
@@ -19,6 +19,7 @@ if [[ ${report//$'\n'/' '} =~ $pattern_lcov ]]; then
         echo "✅ Test coverage rate is higher than 90%"
     else
         echo "❌ Test coverage rate is lower than 90%"
+        echo "You can download detailed coverage report in action's summary"
```
ziobron, line 22: Can you show me where exactly is the link to download the coverage report? Some screenshot + link to some dummy report.

anraf1001: After clicking on `Details` link next to the failed job, you need to go to the `Summary`:
summary
and at the bottom of that site there is an `Artifact` section where you can download the report
artifact
[Example artifact](https://github.com/coders-school/memory-management/suites/6354850517/artifacts/229354561)

PR #157, opened 2022-05-11
Coverage rate comment - unique_ptr

anraf1001:


.gitignore
```diff
@@ -7,7 +7,6 @@ out/
 .svn
 log/*.log
 tmp/**
-node_modules/
```
ziobron, line 10: You should not commit node_modules. Remove them and add to .gitignore.

.github/actions/coverage-rate-comment/postPRInfo.js
```diff
@@ -0,0 +1,27 @@
+const fetch = require('cross-fetch');
+
+async function postPRInfo(owner, repo, prNumber, runId) {
+    const url = 'https://hook.integromat.com/r2dowpjpogyhlt7q825cnk9vc2b741hv';
```
ziobron, line 4: can we somehow hide this address? This may be a security issue. Anyone can send some corrupted data to this webhook and do something nasty, like commenting on other PR :D Can we use Github secrets or something like that?

anraf1001: We can use organization level secret containing that url, but it will cause another problem. In repo level secrets, GitHub says that `Secrets are not passed to workflows that are triggered by a pull request from a fork`.
Maybe there is an option to pass organization's secrets to forks.

anraf1001: @ziobron OK. I have a second idea. We can create a repo for that action and add a repo-level secret containing url. Maybe a local secret dedicated only for that action will work, even if it will be executed from a fork.

ziobron: Let's stay with it right now. Later we will need to fix it.

ziobron:
Your PR was merged!

PR #211, opened 2023-12-06
Problem solved: throwing exception causes skip of delete operator and…

kamilwojcik:
I think this is a good example what can go wrong if we use classic pointers (not smart ones) and we don't pay enough attention to memory management when we handle exceptions.


ziobron:
✅ CI Passed - Task: resource
🏅 Granted 5 XP for @kamilwojcik

